PR #43, opened 2025-06-26
Refactor dependency injection setup 

undead2146:
With this PR I aim to create a clear and modular way to load services into GenHub.

For this I have created a GenHub/Infrastructure/DependencyInjection directory whereby `AppServices.cs` is responsible for adding all the shared services. 

Furthermore I have seperated each service registration and grouped them together into modules. 
`GameDetectionModule.cs` for game detection services.
`SharedViewModelModule.cs` for shared viewmodels.


roossienb:
Please check and fix the stylecop warnings for the three newly added files.

PR #53, opened 2025-07-07
feat(app-update): define core interfaces and models

undead2146:
This PR lays the foundational groundwork by defining all necessary interfaces and data models for the app update feature. This PR includes no concrete implementations, only the public API contracts.

PR #54, opened 2025-07-07
feat(app-update): implement versioning and GitHub services

undead2146:
- Merge after #53

This PR implements the versioning logic (`SemVerComparator`) and the GitHub API client. This PR introduces the `Octokit` dependency and registers the new services in the DI container


roossienb (approved):
Looks good

PR #79, opened 2025-07-24
chore: Clean up .idea directory by removing unnecessary configuration and updated .gitignore

ViTeXFTW:
### Change  
Noticed some JetBrains specific configurations in the GenHub folder. Removed the folder and files from Git tracking and updated `.gitignore` to ignore all files and directories starting with a `.`.

If in the future a `.` directory is needed it will need to be excluded from the general ignore rule in `.gitignore`.


ViTeXFTW:
I updated `.gitignore` so `.vs/` and `.vscode/` is no longer in there as they are covert by the `.*` entry.

PR #149, opened 2025-09-17
CSV EPIC 

mnoserat:


mnoserat (commented):
add undead changes

roossienb:
Additionally, create a clear title what this PR does, add a (short) description what it does and link the relevant issues (`- Resolves #000`). The issue linked in the title is an epic and should not be referred to directly.

GenHub/GenHub.Core/Constants/CsvConstants.cs
```diff
@@ -0,0 +1,142 @@
+using System;
+
+namespace GenHub.Core.Constants;
+
+/// <summary>Constants used for CSV file processing and content handling.</summary>
+public static class CsvConstants
```
RMWMStevens, line 6: Why don't you just use [CsvHelper](https://www.nuget.org/packages/csvhelper/)? It takes care of so much of the CSV work for you, then you don't have to remember what the delimiter is, and if the Language is in column 5, etc. You just read the CSV into an object and you're done. Extremely quick & easy, and also much easier to read code-wise & work with from that point forward.

GenHub/GenHub.Core/Constants/CsvConstants.cs
```diff
@@ -0,0 +1,142 @@
...
+    public const string CsvResolverId = "CSVResolver";
+
+    /// <summary>CSV source name identifier.</summary>
+    public const string CsvSourceName = "CSV";
+
+    /// <summary>File system deliverer source name.</summary>
+    public const string FileSystemSourceName = "FileSystem";
+
+    /// <summary>Default hash algorithm to use for file validation.</summary>
+    public const string DefaultHashAlgorithm = "SHA256";
```
RMWMStevens, line 69: I'm pretty sure this one is already in a constant somewhere.

GenHub/GenHub.Core/Constants/CsvConstants.cs
```diff
@@ -0,0 +1,142 @@
...
+
+    /// <summary>Progress message for CSV download phase.</summary>
+    public const string DownloadingCsvMessage = "Downloading CSV file...";
```
RMWMStevens, line 72: Personally I think if you're only using this message in a log for example, then you really don't have to turn it into a Constant. I think this is overusing constants a little bit.

GenHub/GenHub/Features/Content/Services/ContentDeliverers/CSVDeliverer.cs
```diff
@@ -0,0 +1,291 @@
...
+        try
+        {
+            // Obtain CSV URL from common locations
+            string? csvUrl = null;
+
+            // Try ResolverMetadata (common pattern in other discoverers)
+            try
+            {
+                // Some ContentSearchResult implementations expose a dictionary property named ResolverMetadata
+                // We'll attempt to read it via dynamic/Reflection style if typed property isn't available.
```
RMWMStevens, line 81: So, is the typed property available? It says "if" it's available.
And is there no easier way of doing this than to use reflection? I feel like _often_ it's needlessly complex and reflection isn't very performant.

GenHub/GenHub/Features/Content/Services/ContentDeliverers/CSVDeliverer.cs
```diff
@@ -0,0 +1,291 @@
...
+            {
+                _logger.LogError(CsvConstants.CsvDownloadFailedLog, csvUrl, response.StatusCode);
+                return OperationResult<ContentManifest>.CreateFailure($"{CsvConstants.CsvDownloadFailedError}: {response.StatusCode}");
+            }
+
+            using var stream = await response.Content.ReadAsStreamAsync(cancellationToken);
+            using var reader = new StreamReader(stream);
+
+            // Parse CSV streaming
+            var files = new List<ManifestFile>();
```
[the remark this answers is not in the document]
RMWMStevens: ^ I agree with this one, I already mentioned CsvHelper in another comment.

GenHub/GenHub/Features/Content/Services/ContentDiscoverers/CSVDiscoverer.cs
```diff
@@ -0,0 +1,117 @@
...
+/// CSV URLs and metadata for each supported content type.
+/// </remarks>
+public class CSVDiscoverer(
+    ILogger<CSVDiscoverer> logger,
+    ManifestDiscoveryService manifestDiscoveryService,
+    IConfigurationProviderService configurationProvider) : IContentDiscoverer
+{
+    private readonly ILogger<CSVDiscoverer> _logger = logger;
+    private readonly ManifestDiscoveryService _manifestDiscoveryService = manifestDiscoveryService;
+    private readonly IConfigurationProviderService _configurationProvider = configurationProvider;
```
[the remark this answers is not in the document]
RMWMStevens: Maybe he likes to have all names with underscores, which is not possible directly in the primary constructor. Maybe you can just circumvent this by keeping these lines and adding `?? throw new ArgumentNullException(...)`;

GenHub/GenHub/Features/Content/Services/ContentDiscoverers/CSVDiscoverer.cs
```diff
@@ -0,0 +1,117 @@
...
+        CancellationToken cancellationToken = default)
+    {
+        var results = new List<ContentSearchResult>();
+
+        // Generals 1.08
+        if (query.TargetGame == GameType.Generals)
+        {
+            var searchResult = new ContentSearchResult
+            {
+                Id = "generals-1.08",
```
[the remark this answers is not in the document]
RMWMStevens: Unrelated but it's surprising to me how there is a simple error message in the constants file, but such a value as this ID here is _not_ there.

GenHub/GenHub/Features/Content/Services/ContentProviders/CSVContentProvider.cs
```diff
@@ -0,0 +1,306 @@
...
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Security.Cryptography;
+using System.Threading;
+using System.Threading.Tasks;
+
+// This class implemented in a very basic way and needs more work
+// It needs to be split into smaller methods
+// Also, the CSV format should be more robustly defined (e.g. using a library like CsvHelper)
```
RMWMStevens, line 20: Hey, I'm glad to see you mentioning this yourself! That's makes me happy actually, nice job. I'm curious to see what you come up with.

Maybe for next time though: if something needs more work, consider marking the PR as 'draft' until you're happy with it and have addressed all or most of your own bullet points. Until I saw this section, I was under the impression that this PR is "ready to go", but now I think otherwise and feel like you're saying the same thing here. :)

GenHub/GenHub/Features/Content/Services/ContentProviders/CSVContentProvider.cs
```diff
@@ -0,0 +1,306 @@
...
+            SearchTerm = contentId,
+            Take = ContentConstants.SingleResultQueryLimit,
+        };
+        var searchResult = await SearchAsync(query, cancellationToken);
+        if (!searchResult.Success || !searchResult.Data!.Any())
+        {
+            return OperationResult<ContentManifest>.CreateFailure($"{CsvConstants.ContentNotFoundError}: {contentId}");
+        }
+
+        var result = searchResult.Data!.First();
```
RMWMStevens, line 143: This code is unsafe. Why not write it like this: 

```
var result = searchResult?.Data?.FirstOrDefault();
var manifest = result?.GetData<ContentManifest>();

return manifest != null
    ? OperationResult<ContentManifest>.CreateSuccess(manifest)
    : OperationResult<ContentManifest>.CreateFailure(CsvConstants.ManifestNotAvailableError);
```

This way an exception will be avoided in every single case a variable here could be `null`.

GenHub/GenHub/Features/Content/Services/ContentProviders/CSVContentProvider.cs
```diff
@@ -0,0 +1,306 @@
...
+    {
+        var query = new ContentSearchQuery
+        {
+            SearchTerm = contentId,
+            Take = ContentConstants.SingleResultQueryLimit,
+        };
+        var searchResult = await SearchAsync(query, cancellationToken);
+        if (!searchResult.Success || !searchResult.Data!.Any())
+        {
+            return OperationResult<ContentManifest>.CreateFailure($"{CsvConstants.ContentNotFoundError}: {contentId}");
```
RMWMStevens, line 140: Why not use the `Result.Message` or whatever it's called in the response?

GenHub/GenHub/Features/Content/Services/ContentDiscoverers/CSVDiscoverer.cs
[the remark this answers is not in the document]
roossienb: You would only check for null values if there is a reason for it. We do use `nullable enabled` hence if a parameter isn't explicitly typed as nullable (e.g. `Foo?`) it should not be null (but the compiler will only give a warning, not an error, so there still is a change the object is null).

Not sure which convention @undead2146  normally uses here. If you play it safe, you would add the `ArgumentNullException`.

Then that begs the question of the exception being properly caught higher up.
Now I think about it, this is a wider topic that maybe should be discussed somewhere else.

PR #186, opened 2025-11-23
feat(content-pipeline): add `IContentUpdateService` with base class implementation

undead2146:
This PR introduces a `IContentUpdateService` interface along with a base class for implementations.

This will be used for updating content from ContentProviders that have version system to their content.
I.E. `GeneralOnlineUpdateService` would get the most recent manifests for this publisher.

### DIFF
--- a/GenHub/GenHub.Core/Interfaces/Content/IContentUpdateService.cs
+++ b/GenHub/GenHub.Core/Interfaces/Content/IContentUpdateService.cs
@@ -1,0 +1,14 @@
+namespace GenHub.Core.Interfaces.Content;
+
+/// <summary>
+/// Interface for content update checking services.
+/// </summary>
+public interface IContentUpdateService
+{
+    /// <summary>
+    /// Checks for available content updates.
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token.</param>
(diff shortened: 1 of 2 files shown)


GenHub/GenHub.Core/Interfaces/Content/IContentUpdateService.cs
```diff
@@ -0,0 +1,14 @@
...
+/// Interface for content update checking services.
+/// </summary>
+public interface IContentUpdateService
+{
+    /// <summary>
+    /// Checks for available content updates.
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>A tuple containing update availability, latest version, and current version.</returns>
+    Task<(bool UpdateAvailable, string? LatestVersion, string? CurrentVersion)> CheckForUpdatesAsync(CancellationToken cancellationToken = default);
```
RMWMStevens: I added a comment on the GenHub PR about this, you likely have seen it already.

RMWMStevens (commented):
I feel like I've already seen this. If these changes are also in the Generals Online PR, then I'd recommend removing this PR and just sticking to the Generals Online one.

PR #188, opened 2025-11-23
feat(publisher): implement publisher manifest factory

undead2146:
This PR introduces a `IPublisherManifestFactory` interface that publishers can use to create factory implementations like for example `SuperHackersManifestFactory`, `GeneralsOnlineManifestFactory`, `CommunityOutpostManifestFactory` for generating `ContentManifests`.


RMWMStevens (approved):
TWO files???
That means I'm done reviewing in UNDER an hour! Happy days 🥳
Approved :)

PR #210, opened 2025-12-13
feat(storage): implement dynamic storage location service with user notifications

undead2146:
## Overview
This PR introduces a dynamic storage location system that resolves CAS pool and workspace paths based on game installation locations, enabling hardlink-based workspace creation without requiring administrator privileges. 

### Key Features
- **Path Resolution**: CAS pool and workspace paths are now automatically determined based on the preferred game installation location
- **Hardlink Optimization**: Enables hardlink workspace strategies without admin rights when game and workspace are on the same drive
- **Cross-Drive Support**: Automatically detects and handles multiple game installations across different drives
- **User Notifications**: Real-time feedback during storage setup, launch operations, and strategy fallback scenarios
- **Persistent Preferences**: Remembers user's preferred installation for storage across sessions

### Technical Implementation
- Added `IStorageLocationService` interface with methods for resolving CAS pool and workspace paths
- Implemented `StorageLocationService` that intelligently selects storage locations based on installation drive topology
- Enhanced `UserSettings` with `PreferredStorageInstallationId` and `UseInstallationAdjacentStorage` properties
- Updated `CasConfiguration` to support dynamic path resolution (empty `CasRootPath` triggers dynamic resolution)
- Integrated storage location detection in `GameProfileLauncherViewModel` during game scanning
- Added comprehensive fallback logic in workspace strategies (HardLink, HybridCopySymlink, SymlinkOnly)
- Created `NotificationDurations` constants for consistent notification timeouts
- Added `PathHelper` utility for safe path manipulation
- Enhanced Windows hardlink operations with detailed error reporting via `Win32ErrorCodes`

### User-Facing Changes
During game installation scanning, users will see notifications:
- **Single Drive**: "Storage Location Set" - confirms automatic configuration
- **Multiple Drives**: "Multiple Game Installations Found" - explains auto-selection and potential implications
- **Launch**: Notifications for strategy changes due to missing admin rights
- **Errors**: Clear guidance for cross-drive hardlink failures with actionable solutions

### Configuration
Users can control behavior through settings:
- `UseInstallationAdjacentStorage` (default: `true`) - Enable/disable dynamic storage paths
- `PreferredStorageInstallationId` - Manually set preferred installation for storage

### Breaking Changes
- `ContentStorageService` now requires `storageRoot` parameter in constructor (DI container updated accordingly)
- `GameLauncher` now depends on `IStorageLocationService` instead of `IConfigurationProviderService` for workspace paths
- `CasConfiguration.CasRootPath` defaults to empty string for dynamic resolution (existing configs will continue working)


GenHub/GenHub.Tests/GenHub.Tests.Core/Features/Workspace/TestFileOperationsService.cs
```diff
@@ -0,0 +1,71 @@
...
+    IDownloadService downloadService,
+    ICasService casService) : FileOperationsService(logger, downloadService, casService)
+{
+    private readonly ILogger<FileOperationsService> _logger = logger;
+
+    [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+    private static extern bool CreateHardLink(string lpFileName, string lpExistingFileName, IntPtr lpSecurityAttributes);
+
+    /// <inheritdoc/>
+    public override async Task CreateHardLinkAsync(
```
RMWMStevens: Hard error here

RMWMStevens: Can't you just write this implementation and instead of doing `public override async Task CreateHardLinkAsync`, just inherit the interface instead of the class and make it do nothing? Either that or mark the `CreateHardLinkAsync` method `virtual`, which seems like a bad idea to me if it's only for testing.

GenHub/GenHub/Common/Services/ConfigurationProviderService.cs
```diff
@@ -279,9 +279,38 @@ public string GetContentStoragePath()
     }
 
     /// <inheritdoc />
+    /// <remarks>
+    /// WARNING: If the specific configuration is missing or incomplete, this method may return a
```
RMWMStevens: If it is not missing or incomplete, will the returned instance be "connected from the underlying settings" as opposed to the situation described here? As in, will you then not have to call `TryUpdateAndSaveAsync`?

GenHub/GenHub/Features/GameProfiles/Services/ProfileLauncherFacade.cs
```diff
@@ -61,6 +65,8 @@ public class ProfileLauncherFacade(
     private readonly IDependencyResolver _dependencyResolver = dependencyResolver ?? throw new ArgumentNullException(nameof(dependencyResolver));
```
RMWMStevens: Why all these private fields? Please just use the primary constructor arguments. Unless you have good reason to believe, or have an urgent need to check, if any of these are actually null. These lines don't add any value in their current shape, as far I am concerned. 

I would prefer it if you could check for the use of primary constructors in (new) files, as well as using their arguments without unnecessary local field assignments, next time before marking a PR as 'Open'. I feel like I'm repeating this on every PR.

GenHub/GenHub/Features/GameProfiles/ViewModels/GameProfileLauncherViewModel.cs
```diff
@@ -338,6 +347,105 @@ private async Task ScanForGamesAsync()
...
+            var existingPreferred = await _storageLocationService.GetPreferredInstallationAsync();
+            if (existingPreferred != null)
+            {
+                logger.LogDebug("Preferred storage installation already set: {InstallationId}", existingPreferred.Id);
+                return;
+            }
+
+            // Group installations by drive
+            var installationsByDrive = installations
+                .GroupBy(i => Path.GetPathRoot(i.InstallationPath)?.ToUpperInvariant() ?? string.Empty)
```
RMWMStevens: Wouldn't it make more sense for the `GetPathRoot` method to return `D:\` instead of `d:\`, so the `.ToUpperInvariant()` method be moved there instead? Now if you ever forget, you'll get the `d:\` which looks very strange in drive context.

GenHub/GenHub/Features/Launching/GameLauncher.cs
```diff
@@ -536,40 +548,62 @@ private async Task<LaunchOperationResult<GameLaunchInfo>> LaunchProfileAsync(Gam
...
 
+            // Resolve the installation
+            var installationResult = await gameInstallationService.GetInstallationAsync(profile.GameInstallationId, cancellationToken);
+            if (!installationResult.Success || installationResult.Data == null)
+            {
+                logger.LogError("[GameLauncher] Failed to resolve game installation for profile {ProfileId}", profile.Id);
+                return LaunchOperationResult<GameLaunchInfo>.CreateFailure("Failed to resolve game installation.", launchId, profile.Id);
+            }
+
+            var installation = installationResult.Data!;
```
RMWMStevens: This is such a tiny thing, but I'm 99% sure that this `!` is not necessary here as you already have an `if` before it that specifically checks whether `installationResult.Data == null`. The compiler is smart enough to know that then line 559 will be unreachable if it was `null` :). And let's say someone ever changes the `||` on line 553 to `&&`, then there is a case in which here `.Data` _can_ be null, and due to the `!` it won't show a warning. Hence why I'd prefer to remove the `!`.

GenHub/GenHub/Features/Launching/GameLauncher.cs
```diff
@@ -536,40 +548,62 @@ private async Task<LaunchOperationResult<GameLaunchInfo>> LaunchProfileAsync(Gam
...
+            var dynamicWorkspacePath = storageLocationService.GetWorkspacePath(installation);
+            logger.LogDebug("[GameLauncher] Using dynamic workspace path: {WorkspacePath} (Installation: {InstallPath})", dynamicWorkspacePath, actualInstallationPath);
+
             logger.LogDebug("[GameLauncher] Creating workspace configuration - Strategy: {Strategy}", profile.WorkspaceStrategy);
             var workspaceConfig = new WorkspaceConfiguration
             {
                 Id = profile.Id,
                 Manifests = manifests,
-                GameClient = profile.GameClient!,
+                GameClient = gameClient, // FIX: No null-forgiving operator needed
```
RMWMStevens: This `// FIX:` comment, does it need to stay?

GenHub/GenHub/Features/Launching/GameLauncher.cs
```diff
@@ -752,8 +786,15 @@ private async Task ApplyProfileSettingsToIniOptionsAsync(GameProfile profile, Ca
     {
         try
         {
+            // FIX: Add null check for GameClient to resolve CS8602 warnings
```
RMWMStevens: This is precisely what the below `if` does, correct?

GenHub/GenHub/Features/GameProfiles/ViewModels/GameProfileLauncherViewModel.cs
undead2146: Done, Moved `.ToUpperInvariant(`) to apply directly to the result of `Path.GetPathRoot()`

GenHub/GenHub/Features/Launching/GameLauncher.cs
undead2146: Removed all obsolete // FIX comments.

GenHub/GenHub/Features/Launching/GameLauncher.cs
```diff
@@ -752,8 +786,15 @@ private async Task ApplyProfileSettingsToIniOptionsAsync(GameProfile profile, Ca
...
+            if (profile.GameClient == null)
+            {
+                logger.LogWarning("Profile {ProfileId} has no GameClient configured, skipping Options.ini write", profile.Id);
+                return;
+            }
+
             // Determine the game type from the profile's game client
-            var gameType = profile.GameClient.GameType;
+            var gameType = profile.GameClient.GameType; // Now safe - no CS8602 warning
```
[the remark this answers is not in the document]
undead2146: Done.

RMWMStevens (approved):

PR #222, opened 2025-12-28
feat: fix CD disk detection; add manual folder picker

undead2146:
This PR fixes the GameInstalation not being detected for CDISO whereby users do not have a `Instal Dir` registry entry.

- Fixed game recognition for CD/ISO installations.
- Added manual directory selection fallback.
- Added missing Zero Hour retail paths.
- Updated .gitignore for releases.


GenHub/GenHub.Windows/GameInstallations/WindowsInstallationDetector.cs
```diff
@@ -118,19 +118,66 @@ public Task<DetectionResult<GameInstallation>> DetectInstallationsAsync(Cancella
         return Task.FromResult(result);
     }
 
+    /// <summary>
+    /// Checks if any of the specified executables exist in the given directory.
+    /// </summary>
+    /// <param name="directory">The directory to check.</param>
+    /// <param name="executableNames">The list of executable names to look for.</param>
+    /// <returns>True if any of the executables exist.</returns>
+    private static bool HasAnyExecutable(string directory, string[] executableNames)
```
RMWMStevens: This whole method is basically just this, but much more verbose:

```
private static bool HasAnyExecutable(string directory, string[] executableNames) =>
    executableNames.Any(exe => File.Exists(Path.Combine(directory, exe)));
```

Do you still need a separate method for it? I'd only do so if you use it in multiple places.

GenHub/GenHub/Features/GameProfiles/ViewModels/GameProfileLauncherViewModel.cs
```diff
@@ -1460,4 +1485,124 @@ private async Task<bool> ShowPromptAsync(string title, string message, GameInsta
...
+
+                    logger.LogInformation(
+                        "Created manual Retail installation from parent directory: Generals={HasGenerals}, ZeroHour={HasZeroHour}",
+                        hasGenerals,
+                        hasZeroHour);
+
+                    return installation;
+                }
+            }
+            else
```
RMWMStevens: This `else` block is entirely redundant as you are already doing a `return` from the `if` on line `1570`. I think this could would make more sense / be more readable if the `else` block was removed.

RMWMStevens (approved):

undead2146 (commented):
Addressed the code review comments

RMWMStevens (approved):

PR #329, opened 2026-07-29
feat(macos): add native host, installation detection, and CI

bobtista:
## Summary

Add the native macOS application host, app bundle, installation detection, and CI coverage.

This PR is independent of the cross-platform fixes and can be reviewed and merged separately.

## Changes

- Add the macOS application host, app bundle, and composition root.
- Add macOS installation detection that distinguishes permission denial from a successful empty scan.
- Allow detection to retry after the user grants filesystem access.
- Disable self-update on macOS until signed and notarized artifacts are available.
- Add the macOS CI lane, AVIF platform guard, startup fixes, and composition-root coverage.

## Testing

- `dotnet test GenHub/GenHub.MacOS.slnf -c Release --no-restore`
- GenHub.Tests.Core: 1,341 passed, 0 failed.
- GenHub.Tests.MacOS: 3 passed, 0 failed.

## Risks and rollback

This adds a new platform host without changing Windows or Linux host behavior. Self-update remains deliberately disabled on macOS pending #322.

No persisted-data migration is required, and the changes are revert-safe.

## Related issues

Fixes #319
Part of #327

<details open><summary><h3>Greptile Summary</h3></summary>

Adds the native macOS application host and supporting build coverage.
- Registers macOS-specific installation detection, shortcut behavior, and disabled self-update behavior.
- Packages an unsigned arm64 application bundle and smoke-tests its startup in a dedicated macOS CI lane.
- Adds macOS composition-root and installation-result tests while sharing composition assertions across platform test projects.
- Adjusts shared startup, installation retry, image conversion, and update-service behavior for the new host.
</details>

<details open><summary><h3>Confidence Score: 3/5</h3></summary>

The PR is not yet safe to merge because macOS installation detection still silently omits valid retail trees nested beneath an EA Games directory.

The current detector searches only immediate children of its configured roots and does not include ~/Games, leaving the previously reported copied-retail layout undetectable; the earlier CI action-pinning, missing-script, and bundle-version issues are fixed in the current code.

**Files Needing Attention:** GenHub/GenHub.MacOS/GameInstallations/MacOSInstallationDetector.cs
</details>

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| .github/workflows/ci.yml | Adds a commit-pinned, credential-free macOS build, bundle, startup-smoke-test, test, and artifact lane. |
| .github/scripts/package-macos-app.sh | Creates an unsigned macOS application bundle and now emits validated numeric bundle-version metadata. |
[...]

PR #328, opened 2026-07-29
feat(platform): make paths, manifests, and Unix workspaces platform-aware

bobtista:
## Summary

Make shared GenHub runtime behavior platform-aware across Windows and Unix. This fixes platform-specific settings and data paths, corrects Unix workspace materialization, and introduces deterministic manifest variants needed by native clients.

This PR contains no GenHub.MacOS project changes and can be reviewed independently of the macOS host.

## Changes

- Resolve Options.ini and application-data consumers through the configured platform paths.
- Centralize executable classification and add deterministic platform-variant and entry-point resolution.
- Preserve compatibility with existing flat manifests while rejecting unresolved variant usage.
- Implement Unix hard links with link(2) and make Unix symlink strategies reachable.
- Set execute permissions on workspace-owned copies rather than shared CAS blobs.
- Add behavioral coverage for path conventions, manifest variants, executable classification, and Unix workspace behavior.

## Testing

- `dotnet test GenHub/GenHub.Tests/GenHub.Tests.Core/GenHub.Tests.Core.csproj -c Release --no-restore`
- 1,391 passed, 0 failed.

## Risks and rollback

Unix workspace behavior changes: hard-link strategies now create real hard links instead of silently copying. Executable files are detached before permission changes so shared CAS blobs are not mutated.

Existing flat manifests remain supported. No persisted-data migration is required, and the changes are revert-safe.

## Related issues

Fixes #314
Fixes #316
Part of #327

<details open><summary><h3>Greptile Summary</h3></summary>

The PR centralizes platform-aware manifest and workspace behavior.
- Adds manifest variants, deterministic entry-point resolution, and ingestion gating.
- Adds native Unix hard-link handling and symlink capability registration.
- Moves Unix execute-bit changes onto intended workspace-owned copies.
- Routes application-data and game-settings paths through platform providers.
</details>

<details open><summary><h3>Confidence Score: 2/5</h3></summary>

This PR is not yet safe to merge because most Unix strategies skip executable permission isolation, hard-link permission failures can bypass copy fallback, and writable workspaces can mutate shared CAS content.

The execute-permission fix is only reached by SymlinkOnlyStrategy, EACCES escapes the new hard-link fallback, and intentional hard links leave non-executable CAS objects writable through workspace paths, allowing cross-profile content corruption.

[...]

Issue #327, opened 2026-07-29
Alpha 5: native macOS host and native BGFX client — scope and tracking


bobtista:
## Proposal

Alpha 5 delivers first-class macOS support and launches the community's native BGFX Zero Hour client against the user's retail data — **no Wine**. It lands the four-phase `feat/macos-host` → `feat/native-launch` stack plus the hardening items from its external review.

**Notarization (#322) is not being pursued.** It requires a Developer ID certificate,
which requires a paid Apple Developer Program membership, and that is not being
purchased. #354 fixes the half that needed no certificate: GenHub clears
`com.apple.quarantine` from the game executables it materializes, so the propagation
failure — GenHub opens, the game silently will not launch — no longer happens. What
remains is one documented step for the app bundle itself
(`xattr -dr com.apple.quarantine /Applications/GenHub.app`).

macOS is therefore **installable with one manual step**, not distributable in the
Gatekeeper sense. The signing pipeline is built and verified under an ad-hoc identity
on `feat/macos-signing-notarization`, unmerged, ready if that decision changes.

## Scope

**Definition of done: the three PRs land and #354 lands** — at that point macOS is
installable with the one documented step. Notarization is explicitly out of scope; see
#322. Priority order; "PR" is which stack PR
closes it (see The mac stack below).

| # | Issue | Closed by |
|---|---|---|
| 1 | #314 — Reject multi-variant manifests at ingestion | PR A |
| 2 | #315 — Validate retail archive roots before launch | PR C |
| 3 | #316 — Materialize workspace executables atomically | PR A |
| 4 | #317 — Remove `DYLD_LIBRARY_PATH` / `LD_LIBRARY_PATH` | PR C |
| 5 | #318 — Fix stderr capture EOF race | PR C |
| 6 | #319 — macOS detector must distinguish TCC denial from "none found" | PR B |
| 7 | #320 — `ValidateScopes` shrink-only ratchet | own PR |
| 8 | #322 — signing and notarization | **dropped** — needs a Developer ID; #354 covers the usable-unsigned path |
| 9 | #323 — Launch receipt + cheap revalidation | own PR |
| 10 | #324 — Engine-only launch smoke on `macos-15` CI | own PR |
| 11 | #325 — Magic-byte executable classification | own PR |
| 12 | #330 — Detect installations by archive presence, not executable name | own PR |
| 13 | #333 — Zero Hour can launch with no base Generals content | blocked on engine |
[...]

Issue #324, opened 2026-07-29
Add engine-only launch smoke test on macos-15 CI (needs no retail data)


bobtista:
## Problem

There is no CI coverage of the native launch path. Launch regressions surface only manually, on a developer machine, after the fact.

## Corrected premise — read before implementing

This issue originally said the test "needs no retail data (engine reaches main loop with no content)". **The engine cannot reach its main loop with no content.** Verified in `INI.cpp:222`:

```cpp
// Expect to open and load at least one file.
if (filesRead == 0)
{
    throw INI_CANT_OPEN_FILE;
}
```

`INI_CANT_OPEN_FILE` is in an *unnamed* enum, so it throws a plain int, does not match `catch (ErrorCode ec)`, and falls to `catch (...)` → `RELEASE_CRASH` → `_exit(1)`. All of that is in **upstream** `TheSuperHackers/GeneralsGameCode@main`, not fork-local — observed behaviour, though not design intent: the routing is accidental (`catch (ErrorCode ec)` acts only on `ERROR_INVALID_D3D` and silently swallows anything else, so the plain-int route to `catch (...)` is what actually produces the crash). Assert what is observed, not that it was intended.

So the test as originally specified cannot be built. It is still worth building, asserting the opposite outcome.

## What to assert instead

Launch the engine with **no retail data** and assert it **exits 1** with the initialisation-failure diagnostic. That still validates a great deal, and needs no licensed content:

- the binary loads at all
- dynamic libraries resolve via `@executable_path` / RPATH — the failure mode #317 was about
- initialisation runs far enough to attempt INI loading
- the documented failure path is reachable rather than hanging

A test that reaches the main loop needs content and therefore cannot run in CI. That is a separate, local concern.

## Flags: check fork status before relying on one

| Flag | Upstream `main` | bgfx trunk |
|---|---|---|
| `-headless` | yes | yes |
| `-noDraw`, `-noaudio`, `-quickstart`, `-win` | yes | yes |
| `-perfAutoExitSeconds` | **no** | yes |

Use `-headless` — it exists in both. **Do not build this test on `-perfAutoExitSeconds`**: it is fork-only, so a test depending on it validates our client and breaks against upstream.

`-perfAutoExitSeconds` is still useful *locally*: the three real-client tests currently launch the game and hold it for 14 seconds, which is why they fail whenever a developer closes the window. Self-termination would fix that, and is fine there because those tests only ever run against our client.

## Acceptance criteria

[...]

PR #332, opened 2026-07-30
feat(launching): launch the native BGFX client against retail archives

### LINKED ISSUE #315: Validate retail archive roots before launch
## Problem

When the retail archive roots are wrong, the failure is hard to act on. With no readable INI content the engine aborts during initialisation — exit 1 plus a `ReleaseCrashInfo.txt`, generic and indistinguishable from other init failures — and on Windows `ReleaseCrash` shows a system-modal dialog before exiting, so a host-launched process hangs instead of dying. Nothing in any of that names the misconfigured root, so the user cannot tell what to fix.

*(This issue originally claimed the engine reaches its main loop with no content and no error. That is false — the correction is recorded in #327. The mitigation below stands either way: validating before spawn turns a generic crash into an actionable message.)*

## Proposed approach

Validate the retail archive roots *before* spawning the engine:

- Check for the existence of a sentinel `.big` archive under each configured root.
- Fail the launch with a clear diagnostic naming the offending root, rather than spawning a process that will present an empty game.

## Acceptance criteria

- A launch with a misconfigured or empty retail root fails before spawn with an actionable message.
- A correctly configured root launches unchanged.
- Validation cost is negligible relative to launch (existence check only, no archive parsing).

## Notes

Stack PR 4 (`feat/native-launch`).

The engine does report mount failures (exit 1; on the fork additionally a `[ggc]` stderr sentinel). GenHub-side detection of those post-spawn signals is #334. This issue stands on its own regardless.

### LINKED ISSUE #317: Remove DYLD_LIBRARY_PATH and LD_LIBRARY_PATH from native launch
## Problem

Native launch currently sets `DYLD_LIBRARY_PATH` / `LD_LIBRARY_PATH`. Measurement showed these are **not required** for the engine to resolve its libraries.

Keeping them is actively harmful:

- `dyld` searches `DYLD_LIBRARY_PATH` **before** `@executable_path`, so any same-named library elsewhere on the path silently overrides the intended one. This is a real correctness and supply-chain risk, not a theoretical one.
- Hardened-runtime signing ignores `DYLD_LIBRARY_PATH` anyway, so the variable will stop having any effect the moment GenHub is signed and notarized — meaning current behaviour would change silently at signing time.

## Proposed approach

Remove both variables from the native launch environment and confirm the engine still resolves its libraries via `@executable_path` / `RPATH`.

## Acceptance criteria

- Native launch sets neither variable.
- Engine launches and reaches its main loop on macOS and Linux.
- Behaviour is unchanged before and after hardened-runtime signing.

## Notes

Stack PR 4 (`feat/native-launch`). Related: the notarization issue, where this becomes load-bearing.

### LINKED ISSUE #318: Fix stderr capture EOF race; diagnostics can be truncated exactly when a launch fails
## Problem

The stderr capture path has an EOF race: the timed wait can return before the async output handlers have drained, so captured diagnostics may be truncated or empty exactly when they are most needed — a crashing launch.

## Proposed approach

- Call parameterless `WaitForExit()` **after** the timed wait, which is the documented way to ensure async output handlers have completed.
- Treat `Data == null` as the EOF signal rather than inferring completion from process exit.
- Use a synchronized ring buffer for captured output.
- Retain first-N + last-N lines, so both the startup context and the failure tail survive truncation.
- Bound both byte count and individual line length, so a pathological engine cannot exhaust memory.
- Apply the same drain check to stdout.

## Acceptance criteria

- Diagnostics are complete for a process that exits immediately.
- Capture is bounded in memory under a high-volume or pathological writer.
- Both the head and tail of output are retained when truncation occurs.
- No unbounded growth and no data race on the buffer.

## Notes

Stack PR 4 (`feat/native-launch`).

bobtista:
## Summary

Launches the community's native BGFX Zero Hour client against the user's retail data — no Wine. Final PR of the mac stack, after #328 (platform-neutral paths, manifests and Unix workspaces), #329 (macOS host) and #331 (the macOS service registrations the merged state required).

## Changes

- **Surface native launch failures and set the loader search path.** A launch that dies before drawing a window previously reported success.
- **Reach retail archives via the install-path environment instead of copying them.** The engine reads `CNC_ZH_INSTALLPATH` / `CNC_GENERALS_INSTALLPATH` on non-Windows and mounts `*.big` from those roots. Zero Hour needs its own and the base Generals archives — roughly 3 GB — so without this every profile workspace had to materialise all of them.
- **Validate retail archive roots before spawning.** Checked before spawn so a misconfigured root fails with the path named, rather than as a generic engine abort the host has to interpret.

  *Corrected from this PR's original description, which claimed the engine reaches its main loop with zero content and no error and that nothing was observable. Verified against `bobtista/GeneralsGameCode` at `bobtista/topic/trunk`, that is false: a root holding no archives aborts during initialisation with exit 1 and a `ReleaseCrashInfo.txt`, and an unmountable archive also writes `[ggc] ARCHIVE MOUNT FAILED` to stderr. Neither reaches the main loop.*

  Validating first still earns its place for reasons that hold: exit 1 is generic; the stderr sentinel exists only in `StdBIGFileSystem`, not `Win32BIGFileSystem`; `ReleaseCrash` shows a system-modal dialog before `_exit(1)`, so a host-launched child on Windows hangs rather than dying; and GenHub's own immediate-exit detection is a 500 ms window. An earlier check with an actionable message avoids depending on any of that. Validation runs against the installation's *declared* paths, not the filtered environment — the environment builder drops a path that does not exist, so validating it alone would skip the stale-installation case this exists to catch.
- **Stop setting `DYLD_LIBRARY_PATH` / `LD_LIBRARY_PATH`.** Measured unnecessary: the BGFX build declares every dependency as `@executable_path/…` with a matching rpath and launches with the variable cleared. Two costs outweighed a fallback for a build we do not ship — dyld consults it *before* `@executable_path` for leaf-name references, so a same-named library elsewhere silently wins; and the hardened runtime ignores it, so behaviour would change silently once GenHub is signed and notarized (#322).
- **Capture complete stderr when a launch fails.** The buffer was read as soon as the process exited, but the asynchronous handlers had not necessarily delivered their final lines — truncating diagnostics precisely where they explain the failure. Now drains via the parameterless `WaitForExit()`, which also waits for redirected-output handlers. The buffer treats a null line as end-of-stream rather than content, retains the head as well as the tail so startup context survives alongside the symptom, and bounds line length and total size.

## Testing

- `GenHub.Tests.Core`: **1427 passed**, 0 failed.
- `GenHub.Tests.MacOS` composition root: 3 passed.
- Builds with **0 warnings**.

### The launch tests pass locally and skip in CI — please read this before trusting a green tick

Six tests exercise the real client: three launch it and assert it is still alive after 14 seconds, three assert a misconfigured launch does *not* survive. All six pass on a machine with a native install (`GENHUB_NATIVE_CLIENT_DIR`, 6/6 in 54s).

**They skip in CI**, which has no native fixture and no retail data. A green CI run on this PR therefore says nothing about whether launching works — it only covers the unit and composition tests. The engine-only CI smoke test that would close this gap is #324, and it is deliberately out of scope here.

## Fork dependency worth stating

The install-path mechanism this PR validates is **fork-only**. `CNC_ZH_INSTALLPATH` and
`CNC_GENERALS_INSTALLPATH` are read by `bobtista/GeneralsGameCode` on `bobtista/topic/trunk`
and do not exist in `TheSuperHackers/GeneralsGameCode@main` — upstream resolves install paths
from the Windows registry only.

That is fine, since GenHub launches our client, but it bounds what the validation means:
against an upstream build these variables are ignored and the roots have no effect. The
constants and their rationale live in `RetailArchiveConstants` so this is discoverable rather
than implicit.

The engine behaviour the validation *complements* is upstream, though — the `INI`
`filesRead == 0` throw, the unnamed enum that routes it to `catch (...)`, and `ReleaseCrash`
showing a system-modal dialog before `_exit(1)` are all in upstream `main`. So the reasons
for checking before spawn rather than relying on the child's exit hold for any build.

## Risks and rollback

Revert-safe; no persisted-data conversion. The archive-root validation stops launches before spawn that would previously have failed only after spawn, with a generic engine crash naming nothing — intended, and the validation message names the offending root.

## Related

Part of Alpha 5 (#327).

Closes #315.
Closes #317.
Closes #318.

*(One keyword per line: GitHub links only the first reference in `Closes #315, #317, #318`, which would have left #317 and #318 open after merge.)*

This PR enables native BGFX clients to launch against retail archive roots and improves startup diagnostics.
- Centralizes the retail archive environment-variable contract and normalizes configured roots.
- Validates applicable archive roots before spawning native clients, with Windows excluded from the non-Windows contract.
- Captures bounded stderr output and reports immediate process failures.
- Adds unit and opt-in native integration coverage for archive roots, manifests, permissions, and diagnostics.
</details>

<details open><summary><h3>Confidence Score: 3/5</h3></summary>

The PR is not yet safe to merge because native Zero Hour can still be spawned without any validated source of its required base Generals archives.

The current validation skips an absent GeneralsPath, while supported engine-only workspaces contain no archives and rely entirely on environment-provided retail roots; that leaves the previously reported missing-base-content startup failure reachable.

**Files Needing Attention:** GenHub/GenHub/Features/Launching/GameLauncher.cs; GenHub/GenHub.Tests/GenHub.Tests.Core/Features/Launching/RetailArchiveRootValidationTests.cs
</details>

| Filename | Overview |
|----------|----------|
| GenHub/GenHub/Features/Launching/GameLauncher.cs | Builds and validates native retail archive environments; the previous missing-Generals-root failure remains reachable. |
| GenHub/GenHub/Features/GameProfiles/Infrastructure/GameProcessManager.cs | Adds Unix execute-permission checks, bounded asynchronous stderr capture, and clearer immediate-exit handling. |
| GenHub/GenHub.Core/Constants/RetailArchiveConstants.cs | Centralizes archive environment names and case-insensitive top-level archive enumeration settings. |
| GenHub/GenHub.Tests/GenHub.Tests.Core/Features/Launching/RetailArchiveRootValidationTests.cs | Covers archive-root validation behavior but codifies acceptance of a Zero Hour launch with no Generals root. |
| GenHub/GenHub.Tests/GenHub.Tests.Core/Features/GameProfiles/RetailArchiveRootTests.cs | Adds opt-in native coverage proving engine-only workspaces rely on both retail archive environment roots. |

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Launcher as GameLauncher
    participant Install as GameInstallation
    participant Process as GameProcessManager
    participant Engine as Native BGFX Client
    Launcher->>Install: Read Zero Hour / Generals roots
    Launcher->>Launcher: Build normalized environment
    Launcher->>Launcher: Validate declared archive roots
    alt Validation fails
        Launcher-->>Launcher: Return actionable launch error
    else Validation passes
        Launcher->>Process: StartProcessAsync(configuration)
        Process->>Engine: Spawn with redirected stderr
        alt Immediate exit
            Process->>Process: Drain and bound stderr
            Process-->>Launcher: Return failure diagnostics
        else Process survives startup window
            Process-->>Launcher: Return tracked process
        end
    end
```
</details>

<sub>Reviews (10): Last reviewed commit: ["chore(launching): reference launch const..."](https://github.com/community-outpost/genhub/commit/d1adeeb570f233e4a2bec48b9512adf6766e2cd9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48562981)</sub>

**Context used:**

- Knowledge Base — [Game Profiles and Launching](https://app.greptile.com/genhub/-/custom-context/knowledge-base/community-outpost/genhub/-/docs/game-profiles-launching.md)

### DIFF
--- a/GenHub/GenHub.Core/Constants/RetailArchiveConstants.cs
+++ b/GenHub/GenHub.Core/Constants/RetailArchiveConstants.cs
@@ -1,0 +1,42 @@
+namespace GenHub.Core.Constants;
+
+/// <summary>
+/// The contract between GenHub and a non-Windows engine build for locating retail archives.
+/// </summary>
+/// <remarks>
+/// A non-Windows engine reads <c>InstallPath</c> through <c>GetStringFromRegistry</c>, which
+/// on these platforms consults these variables before anything else, then mounts
+/// <c>*.big</c> from those roots in addition to the working directory. The names are
+/// therefore an external contract: changing one silently stops the engine finding content,
+/// with no error from either side.
+/// <para>
+/// Windows resolves install paths from the registry and does not read these, so nothing is
+/// set — and nothing should be validated against them — on that platform.
+/// </para>
+/// </remarks>
+public static class RetailArchiveConstants
+{
+    /// <summary>Environment variable naming the Zero Hour retail directory.</summary>
+    public const string ZeroHourInstallPathVariable = "CNC_ZH_INSTALLPATH";
+
+    /// <summary>Environment variable naming the Generals retail directory.</summary>
+    public const string GeneralsInstallPathVariable = "CNC_GENERALS_INSTALLPATH";
+
+    /// <summary>
+    /// Search pattern for the archives the engine mounts from a retail root.
+    /// </summary>
+    /// <remarks>
+    /// Used as an existence sentinel rather than matching a specific filename, which varies
+    /// by localisation and version.
+    /// </remarks>
+    public const string ArchiveSearchPattern = "*.big";
+
+    /// <summary>
+    /// Every retail archive root variable, Zero Hour first.
+    /// </summary>
+    public static readonly string[] InstallPathVariables =
+    [
+        ZeroHourInstallPathVariable,
+        GeneralsInstallPathVariable,
+    ];
+}
--- a/GenHub/GenHub.Core/Constants/RetailArchiveConstants.cs
+++ b/GenHub/GenHub.Core/Constants/RetailArchiveConstants.cs
@@ -1,3 +1,5 @@
+using System.IO;
+
 namespace GenHub.Core.Constants;
 
 /// <summary>
@@ -30,6 +32,28 @@ public static class RetailArchiveConstants
     /// by localisation and version.
     /// </remarks>
     public const string ArchiveSearchPattern = "*.big";
+
+    /// <summary>
+    /// How <see cref="ArchiveSearchPattern"/> is matched within a retail root.
+    /// </summary>
+    /// <remarks>
+    /// Case-insensitive because retail data copied from a disc or a Windows machine is
+    /// frequently upper-cased, while the default glob is case-sensitive on Linux volumes and on
+    /// case-sensitive APFS. <c>INIZH.BIG</c> would otherwise read as no archives at all and
+    /// block a launch that would have worked — the opposite of what the check exists to do.
+    /// <para>
+    /// <see cref="EnumerationOptions.IgnoreInaccessible"/> is set back to <c>false</c> because
+    /// it defaults to <c>true</c> here, unlike the <see cref="SearchOption"/> overload. Left at
+    /// the default it turns an unreadable root into "no archives found", reporting a permission
+    /// problem as missing content.
+    /// </para>
+    /// </remarks>
+    public static readonly EnumerationOptions ArchiveSearch = new()
+    {
+        MatchCasing = MatchCasing.CaseInsensitive,
+        RecurseSubdirectories = false,
+        IgnoreInaccessible = false,
+    };
 
     /// <summary>
     /// Every retail archive root variable, Zero Hour first.
--- a/GenHub/GenHub.Tests/GenHub.Tests.Core/Features/GameProfiles/BoundedErrorBufferTests.cs
+++ b/GenHub/GenHub.Tests/GenHub.Tests.Core/Features/GameProfiles/BoundedErrorBufferTests.cs
@@ -1,0 +1,104 @@
+using System.Reflection;
+using GenHub.Features.GameProfiles.Infrastructure;
+using Xunit;
+
+namespace GenHub.Tests.Core.Features.GameProfiles;
+
+/// <summary>
+/// Tests for the bounded stderr capture used to explain a failed launch.
+/// </summary>
+/// <remarks>
+/// The buffer is a private nested type; it is exercised through reflection rather than
+/// being made public, because it is an implementation detail of process management and
+/// its behaviour only matters through the diagnostics it produces.
+/// </remarks>
+public class BoundedErrorBufferTests
+{
+    private static readonly Type BufferType =
+        typeof(GameProcessManager).GetNestedType("BoundedErrorBuffer", BindingFlags.NonPublic)!;
+
+    /// <summary>
+    /// The startup context is where the cause usually is, so the first lines must survive
+    /// even when far more output follows than the buffer retains.
+    /// </summary>
+    [Fact]
+    public void Append_RetainsBothTheHeadAndTheTail()
+    {
+        var buffer = CreateBuffer();
+
+        Append(buffer, "dyld: library not loaded");
+        for (var i = 0; i < 200; i++)
+        {
+            Append(buffer, $"noise line {i}");
+        }
+
+        Append(buffer, "Abort trap: 6");
+
+        var text = buffer.ToString()!;
+
+        Assert.Contains("dyld: library not loaded", text);
+        Assert.Contains("Abort trap: 6", text);
+        Assert.Contains("omitted", text);
+    }
+
+    /// <summary>
+    /// A single pathological line must not be retained in full.
+    /// </summary>
+    [Fact]
+    public void Append_TruncatesAnOverlongLine()
+    {
+        var buffer = CreateBuffer();
+
+        Append(buffer, new string('x', 10_000));
+
+        var text = buffer.ToString()!;
+
+        Assert.Contains("line truncated", text);
+        Assert.True(text.Length < 10_000, "The overlong line was retained in full.");
+    }
+
+    /// <summary>
+    /// A null line is the framework's end-of-stream signal, not content.
+    /// </summary>
+    [Fact]
+    public void Append_TreatsNullAsEndOfStreamRatherThanContent()
+    {
+        var buffer = CreateBuffer();
+
+        Assert.False(EndOfStreamReached(buffer));
+
+        Append(buffer, "something failed");
+        Append(buffer, null);
+
+        Assert.True(EndOfStreamReached(buffer));
+        Assert.Equal("something failed", buffer.ToString());
+    }
+
+    /// <summary>
+    /// Total retained output stays bounded regardless of how much arrives.
+    /// </summary>
+    [Fact]
+    public void Append_BoundsTotalRetainedOutput()
+    {
+        var buffer = CreateBuffer();
+
+        for (var i = 0; i < 5_000; i++)
+        {
+            Append(buffer, new string('y', 500));
+        }
+
+        Assert.True(
+            buffer.ToString()!.Length < 128 * 1024,
+            "Retained output grew beyond the cap.");
+    }
+
+    private static object CreateBuffer() => Activator.CreateInstance(BufferType, nonPublic: true)!;
+
+    private static void Append(object buffer, string? line) =>
+        BufferType.GetMethod("Append", BindingFlags.NonPublic | BindingFlags.Instance)!
+            .Invoke(buffer, [line]);
+
+    private static bool EndOfStreamReached(object buffer) =>
+        (bool)BufferType.GetProperty("EndOfStreamReached", BindingFlags.NonPublic | BindingFlags.Instance)!
+            .GetValue(buffer)!;
+}
--- a/GenHub/GenHub.Tests/GenHub.Tests.Core/Features/GameProfiles/NativeClientFixture.cs
+++ b/GenHub/GenHub.Tests/GenHub.Tests.Core/Features/GameProfiles/NativeClientFixture.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System.Runtime.InteropServices;
 
 namespace GenHub.Tests.Core.Features.GameProfiles;
@@ -33,7 +34,10 @@ public static class NativeClientFixture
             var configured = Environment.GetEnvironmentVariable(EnvironmentOverride);
             if (!string.IsNullOrWhiteSpace(configured))
             {
-                return System.IO.Directory.Exists(configured) ? configured : null;
+                // Validated the same way as the discovered default below. A directory that
+                // exists but holds no engine binary is a misconfigured override, and these
+                // tests should skip rather than fail on it.
+                return File.Exists(Path.Combine(configured, BinaryName)) ? configured : null;
             }
 
             var home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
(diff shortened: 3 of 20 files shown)


bobtista:
Addressed the Greptile findings. Both P1s were real and both were introduced by the archive-root validation change in this PR, so thanks — these would have shipped.

**P1: Windows archive validation blocks launches.** Confirmed. `BuildEnvironmentVariables` returns before setting the variables on Windows, but the validation reads the installation's *declared* paths rather than the built environment — deliberately, to catch a stale root that the environment builder silently drops — so it did not inherit that guard. On Windows, where install paths come from the registry and these variables are never read, a layout without loose top-level archives would have been rejected with a misleading message. Now guarded explicitly, with the reason recorded at the guard.

**P1: profile archive roots remain unnormalized.** Confirmed, and it defeated the purpose of the check. `AddArchiveRoot` returned early when the profile had already set the variable, so it never appended the separator; validation then accepted the value because it only tests existence and archive presence. The engine concatenates the value with the archive filename directly, so `/path/to` became `/path/toINIZH.big` and mounted nothing — the exact silent failure this PR exists to prevent, reachable through the override path. A profile override now chooses the directory but not whether the separator is applied.

**P2: centralize archive environment constants.** Agreed — these names are an external contract with the engine, not local strings. Moved to `RetailArchiveConstants` with the `*.big` search pattern, and the rationale for why Windows is excluded lives there too.

Verification: the normalization test fails against the previous early-return, so it pins the fix rather than merely passing. 1423 unit tests pass, 0 warnings. The Windows guard asserts only on Windows and is a no-op elsewhere, so CI's Windows job is what exercises it.

bobtista:
@coderabbitai review

Your previous approval was dismissed by a push, and the re-review hit a rate limit. Since then:

- Fixed the three Greptile findings (two P1s, both real and both introduced by the archive-root validation in this PR) — commit `bdb0b19`.
- Fixed a Windows test regression that `Build Windows` caught: adding the platform guard to validation made two tests contradict it, since they asserted cross-platform rejection — commit `5ac6b3a`.

All checks are green: Build Linux, Build Windows, Build macOS, Build Summary, Detect File Changes, Greptile and Kilo.

bobtista:
@coderabbitai full review

The incremental review reports the commits as already seen, so no new review is produced and the approval dismissed by the push is not restored. Requesting a full re-review of `5ac6b3a` so the approval state reflects the current head.

For context, the two commits since your approval of `bdb0b19`:
- `bdb0b19` — the three Greptile findings (Windows validation guard, profile archive-root normalization, constants extraction)
- `5ac6b3a` — scoping two archive-root rejection tests to non-Windows, which `Build Windows` caught after the guard was added

All checks are green.

bobtista:
Addressed all five findings from the full review. Four were real; one I widened.

**Validate only the launching game's root** — real, and an over-reach I introduced. Validation checked both declared paths regardless of which game was launching, so a stale Zero Hour root could block a Generals launch that would have worked fine. Now takes the game type from the client and checks only that root, keeping the Windows guard.

**Zero-exit fallback dropped the captured stderr** — real. A process exiting immediately with code 0 and no spawned child returned a bare message while the stderr drained moments earlier went unreported. That was the only diagnostic available for the case, and contrary to the point of the capture work in this branch.

**Duplicated fixture discovery** — real, and in three classes rather than two: `RetailArchiveRootTests` had it too. Extracted to `NativeClientFixture` in its own file, so they cannot diverge on when to skip. Divergence there would be quiet and nasty: some would skip while others failed.

**`.so` missing from engine-only staging** — real. The filter matched the binary and `.dylib` only, so the staged workspace was incomplete on Linux.

**StyleCop ordering in the validation tests** — real, and it led somewhere worse. Fixing it surfaced that `GenHub.Tests.Core` had warnings I had never seen, because every "0 warnings" check this branch reported was run against `GenHub.csproj` alone. Five of them had already reached `development` via #328. All are now cleared, including those five, along with the platform-analyzer warnings on `SetUnixFileMode`.

Verification: `GenHub` 0 warnings, `GenHub.Tests.Core` 0 warnings, 1420 unit tests pass, macOS composition 3/3. Ran with the real-client launch tests excluded — they launch the actual game on the developer's machine, so they are only run deliberately.

bobtista:
Greptile's new P1 — "Generals archives escape validation" — was right, and it contradicted CodeRabbit's finding from the same round. Neither stated the full rule, so I reconciled them rather than picking a side.

CodeRabbit wanted validation scoped to the launching game, which is correct for Generals: it does not read the Zero Hour root, so a stale one there must not fail its launch. But Zero Hour is an expansion and mounts the base Generals archives too — the remarks on `AddRetailArchiveRoots` in this file say exactly that, "Zero Hour needs both its own and the base Generals archives" — so scoping it to one root reintroduced the silent failure one directory over: a stale Generals root would let Zero Hour start without base content.

The rule now:

| Launching | Roots validated |
|---|---|
| Generals | its own only |
| Zero Hour | its own, plus the Generals root **when declared** |

The Generals root is checked for Zero Hour only when declared, because an installation carrying the base archives itself needs no separate root — an absent one is not an error, a declared-but-broken one is.

Three tests cover it: a broken declared Generals root fails a Zero Hour launch, an undeclared one does not, and a stale Zero Hour root does not affect a Generals launch. The first fails if the second root is dropped, so it pins the behaviour.

`GenHub` 0 warnings, `GenHub.Tests.Core` 0 warnings, 1423 unit tests, macOS composition 3/3.

I know this dismisses the approval that had just landed — the finding was a genuine correctness gap, so it seemed worth the extra round.

bobtista:
On the "Generals archives escape validation" finding — you were right and my first fix was incomplete. You listed three conditions and I only handled two.

**Fixed in `aeaa1c0`:** a declared Generals root that is stale, and one that holds no archives, now both fail a Zero Hour launch. Generals launches still ignore the Zero Hour root, since they do not read it.

**Not fixed, and it cannot be from here: the absent case.** An absent Generals root is not by itself wrong — the engine mounts archives from the working directory as well, so base content may legitimately sit in the profile workspace rather than a retail root. That is the arrangement the install-path environment replaces, and it remains valid, so failing whenever no root is declared would break it.

Telling the two apart needs to know whether base content is actually reachable, and that is not testable from filenames. A real install holds arbitrary archives beside the retail ones:

```
!!!HotkeyIDMarkersZHEnglish_v2.big
310_ExpandedLANLobbyMenu.big
340_ControlBarPro720ZH.big
```

So `*.big` presence proves nothing about base content specifically. I considered a workspace `*.big` check and rejected it: a Zero Hour workspace always contains archives, so it would always pass — worse than no check, because it reads as a guard while guaranteeing nothing.

Closing it properly needs the engine to report a failed mount, which is the same signal #315 already notes it needs for the archive-exists-but-fails-to-mount case.

Actions taken rather than leaving it implicit:
- `7e6f8ce` records the gap in the code at the decision point, and corrects the comment there. The previous version claimed a Zero Hour installation may carry the base archives itself — I had asserted that without verifying it, and it is not the reason.
- Filed #333 under Alpha 5 so the gap survives #315 being closed by this PR, with the engine-side signal as its acceptance criterion.

bobtista:
All three addressed in `69460ea`. All were real, all in the integration fixtures.

**Fixture override not validated like the default.** `GENHUB_NATIVE_CLIENT_DIR` only checked the directory existed, while the discovered default also required the engine binary. A directory without one is a misconfigured override, so these tests should skip rather than fail — both paths now validate identically. I introduced this asymmetry when extracting the shared helper: the original had it, and I carried it across without noticing.

**Versioned Linux libraries missed.** `.so` as a suffix test misses `libSDL3.so.0`, which is the common shape of a shipped Linux library. Worth noting the second half of this: a copy of the same filter in `NativeClientManifestLaunchTests` had never been updated when `.so` was added earlier, so my previous "`.so` fixed" was only half done. Both now route through a shared `NativeClientFixture.IsDynamicLibrary`.

**Weak manifest assertion.** It required only something executable, so it would have passed with just the binary and never exercised library handling. It now requires a captured dynamic library.

I unit-tested the predicate, because every test that consumes it skips without a local engine install and it would otherwise never execute in CI. That paid for itself immediately — the first implementation matched on `".so"`, which found the coincidental occurrence inside `".sound"` and stopped searching, so `mylib.sound.so.1` was rejected. It matches on `".so."` now and that case is pinned as a test.

`GenHub` 0 warnings, `GenHub.Tests.Core` 0 warnings, 1435 unit tests pass.

bobtista:
Recording decisions settled during the 2026-07-30 fork/upstream investigation, so review does not re-open them:

- **The `CNC_*` install-path dependency is accepted as-is.** Upstream cannot produce a native Unix game binary (its game targets are Windows-only; the advertised "Linux" build cross-compiles the Windows exe for Wine), so no configuration exists where GenHub sets these variables and a real client ignores them — the only native Unix client is the fork, which reads them. On Windows GenHub never sets them and every build resolves installs from the registry, so the stock-retail baseline is unaffected.
- **No client-type detection or branching.** One native target exists, and the Windows path already differs by construction.
- **Exact-name `generalszh` matching stays.** `RTS_BUILD_OUTPUT_SUFFIX` is an empty-by-default CMake cache variable that no preset or CI job sets on either upstream or the fork, applied uniformly to all configs; a suffixed binary requires a manual `-D` override. The manifest `EntryPoint` remains the contract.
- **`-Fix` and `-installation` are not engine flags** (nor GenHub CLI arguments). One is a content-filename fragment, the other a manifest-ID substring; neither reaches any command line, and both engines' flag tables lack them.
- **Trailing-separator normalisation stays for now**, but the underlying concatenation is a fork-side engine bug (upstream's equivalent code is separator-safe and never instantiated), tracked for the engine repo; when fixed there, `EnsureTrailingSeparator` and its contract can be retired.

undead2146 (approved):
LGTM

PR #353, opened 2026-08-03
test(launching): add engine-only launch smoke test and macos-15 CI job

### LINKED ISSUE #324: Add engine-only launch smoke test on macos-15 CI (needs no retail data)

bobtista:
## Summary

Adds an engine-only launch smoke test and a `macos-15` CI job, and makes the build
summary report job outcomes accurately.

Closes #324.

## Why this needs no retail data

The test asserts the engine's documented failure path rather than a successful
launch. With no readable content the engine aborts during initialisation with exit 1
and writes `ReleaseCrashInfo.txt` — the same property #315 relies on for pre-spawn
validation. Asserting that abort exercises the real binary end to end without
shipping or fixturing any retail archives.

The test pins the retail install-path variables, asserts the crash report's reason
line, and confirms the report lands under the sandboxed `HOME` so CI leaves nothing
behind.

## Build summary changes

Two related fixes, both in `ci.yml`:

- `engine-launch-smoke` is included in the summary's `needs`, so its result is
  reflected rather than ignored.
- Failed jobs now fail the summary, and skipped jobs render as skipped instead of
  as "❌ Failed". A skipped `build-macos` previously reported as a failure, which
  made a green run look broken.

## Rebase note

This branch was stacked on `feat/native-launch`, which squash-merged as #332. Its
five commits were replayed onto `development` with
`git rebase --onto origin/development d1adeeb`. No conflicts and no adaptation
needed.

## Testing

- `dotnet test GenHub/GenHub.Tests/GenHub.Tests.Core/GenHub.Tests.Core.csproj -c Release`
  - Passed: 1,538
  - Failed: 0
- The `macos-15` job itself is exercised by this PR's own CI run.

Adds a macOS engine-only launch smoke test and updates the Build Summary to account for all prerequisite jobs accurately.
- Downloads and stages the ARM64 native engine on `macos-15`.
- Exercises the engine’s expected no-content abort and validates its crash report.
- Adds `detect-changes` and the smoke job to the summary dependencies and propagates failed or cancelled results.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

The PR appears safe to merge.

No blocking failure remains; the summary now directly observes `detect-changes`, runs regardless of prerequisite outcomes, and fails when any required job fails or is cancelled.
</details>

| Filename | Overview |
|----------|----------|
| .github/workflows/ci.yml | Adds the macOS smoke job and correctly closes the previously reported gate hole by including `detect-changes` directly in the summary’s dependencies. |
| GenHub/GenHub.Tests/GenHub.Tests.Core/Features/GameProfiles/EngineLaunchSmokeTests.cs | Adds an engine-only smoke test that launches in an isolated workspace and verifies the documented no-content failure path. |

</details>

<sub>Reviews (4): Last reviewed commit: ["ci: drop the unused checkout from the bu..."](https://github.com/community-outpost/genhub/commit/caeadd60b0455cd831d099436e48e3e48d8af0d9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49662145)</sub>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -535,6 +535,66 @@ jobs:
           if-no-files-found: ignore
           retention-days: 7
 
+  # Launches the real native engine with no game data at all and requires the abort the
+  # engine is known to produce (exit code 1 once INI loading finds nothing to read). No
+  # licensed retail content is involved; what this covers is everything before that
+  # point — the binary loads, its dylibs resolve, and startup fails fast instead of
+  # hanging. No other job executes the native launch path at all.
+  engine-launch-smoke:
+    name: Engine Launch Smoke Test
+    needs: detect-changes
+    if: ${{ github.event_name == 'workflow_dispatch' || needs.detect-changes.outputs.any == 'true' }}
+    # macos-14 and newer are Apple Silicon, matching the arm64 engine asset.
+    runs-on: macos-15
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          persist-credentials: false
+
+      - name: Setup .NET
+        uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4
+        with:
+          dotnet-version: ${{ env.DOTNET_VERSION }}
+
+      - name: Cache NuGet Packages
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+        with:
+          path: ~/.nuget/packages
+          key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.csproj') }}
+          restore-keys: |
+            ${{ runner.os }}-nuget-
+
+      # `latest-bgfx` is a moving tag, republished as the engine advances. Deliberate
+      # tradeoff: this job tracks the current engine build rather than pinning a
+      # reproducible one, so an engine regression surfaces here first.
+      - name: Download Native Engine
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh release download latest-bgfx \
+            --repo bobtista/GeneralsGameCode \
+            --pattern 'GeneralsZH-macos-arm64.zip' \
+            --dir engine-download
+          unzip -q engine-download/GeneralsZH-macos-arm64.zip -d native-client
+          chmod +x native-client/generalszh
+          ls -l native-client
+
+      # GENHUB_REQUIRE_NATIVE_SMOKE turns "no client found" from a silent skip into a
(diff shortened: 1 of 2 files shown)


undead2146 (dismissed):
LGTM

PR #247, opened 2026-01-08
feat: implement Tool Profile support for standalone executables and modding utilities

undead2146:
<h2>Greptile Overview</h2>

## Overview

This PR introduces **Tool Profiles**, a specialized classification of `GameProfile` designed for launching standalone executables (specifically the new `ModdingTool` content type). This allows users to manage utilities like WorldBuilder or community tools within the same profile system, but with an accelerated execution path that bypasses game installation requirements.

### Core Architecture Changes
- **New Content Type**: Added `ContentType.ModdingTool` to represent standalone tools and utilities
- **Model Updates**: `GameProfile` now includes `ToolContentId` property and computed `IsToolProfile` property
- **Centralized Validation**: Created `ToolProfileHelper` with both async and sync validation methods for Tool Profile detection
- **Constants Class**: New `ProfileValidationConstants` follows project standards for validation messages

### Launch Pipeline Enhancement
The `ProfileLauncherFacade` implements a bifurcated launch path:
- **Tool Profile Path**: Direct executable launch with optional workspace hydration from CAS, bypassing game installation validation
- **Regular Profile Path**: Existing full workspace assembly and game-specific preparation logic
- **Auto-Detection**: Implicit Tool Profile detection and conversion for profiles with single ModdingTool content

### Content Management
- **Publisher Reclassification**: SuperHackers content reclassified from `GameClient` to `ModdingTool` to enable standalone launch
- **Source Path Mapping**: `ContentStorageService` creates `source.path` mapping files for metadata-only content, enabling physical path resolution
- **File Inclusion Logic**: Expanded SuperHackers manifest factory to include critical dependencies (.dll, .big, .ini files) for tool execution

### UI/UX Updates
- **Dynamic Filters**: `GameProfileSettingsViewModel` implements dynamic content type filtering with ModdingTool support
- **Settings Bypass**: `GameSettingsViewModel` intelligently disables game settings for Tool Profiles
- **Visual Identity**: Added dedicated Tools icon (cog/settings SVG) to content filters

### Documentation
All relevant documentation updated to reflect Tool Profile architecture, including flowcharts, architecture diagrams, manifest ID system, and onboarding guides.

## Notable Design Decisions

1. **Workspace Hydration**: Tool Profiles can use WorkspaceManager for CAS content hydration into `AppData/GenHub/ToolWorkspaces`, respecting user workspace strategies
2. **Validation Bypass**: Tool Profiles skip GameInstallation and GameClient requirements but enforce single-tool constraint
[...]

PR #356, opened 2026-08-07
fix(settings): write the declared TheSuperHackers defaults to settings.json

bobtista:
## Problem

`GameSettingsMapper.ApplyToGeneralsOnlineSettings` fell back to hardcoded literals that disagreed with both the declared model defaults and the game engine. `GameLauncher.ApplyGeneralsOnlineSettingsAsync` runs on every Zero Hour launch and builds a **fresh** `GeneralsOnlineSettings` before mapping, so any profile that never set these got the wrong value written to settings.json on every launch.

Eight settings were affected. Verified against the engine source:

| Setting | Engine | Was | Now |
|---|---|---|---|
| `SystemTimeFontSize` | `8` | `?? 12` | `8` |
| `NetworkLatencyFontSize` | `8` | `?? 12` | `8` |
| `RenderFpsFontSize` | `8` | `?? 12` | `8` |
| `PlayerObserverEnabled` | `TRUE` | `?? false` | `true` |
| `CursorCaptureEnabledInFullscreenGame` | in default mask | `?? false` | `true` |
| `CursorCaptureEnabledInFullscreenMenu` | in default mask | `?? false` | `true` |
| `CursorCaptureEnabledInWindowedGame` | in default mask | `?? false` | `true` |
| `ScreenEdgeScrollEnabledInFullscreenApp` | in default mask | `?? false` | `true` |

Engine references — `GlobalData.cpp` for the font sizes, `OptionPreferences.cpp` for the observer fallback, and the bitmask constants these resolve through:

```cpp
m_systemTimeFontSize = 8;  m_renderFpsFontSize = 8;  m_networkLatencyFontSize = 8;

CursorCaptureMode_Default = EnabledInWindowedGame | EnabledInFullscreenGame | EnabledInFullscreenMenu
ScreenEdgeScrollMode_Default = ScreenEdgeScrollMode_EnabledInFullscreenApp
```

`CursorCaptureEnabledInWindowedMenu` and `ScreenEdgeScrollEnabledInWindowedApp` are absent from those masks, so their existing `?? false` was already correct — they are now expressed through constants for consistency rather than changed.

The cursor capture and edge scroll entries are input behaviour, not cosmetics: every launch was disabling cursor capture and fullscreen edge scrolling for profiles that never touched them.

## Cause

#208 established the correct defaults across the model, constants and view model. #227 bumped the **view model** font sizes to `12`. #247 then added this mapper using `?? 12` and `?? false` — matching the drifted view model rather than the model defaults, despite the block comment saying "use null-coalescing with model defaults". Nothing tested any of it, so the drift went unnoticed.

## Change

- Mapper and `GameSettingsViewModel` both source every TSH default from `GameSettingsTheSuperHackersConstants`, so the UI cannot offer one value while the launcher writes another.
- Boolean defaults added to that constants class, each documented with the engine construct it derives from.
- `ResolutionFontAdjustment` switched to its constant; same value (`-100`), no behaviour change.
- Tests cover fallbacks-match-constants, fallbacks-match-model, and explicit-values-pass-through, for both the numeric and boolean groups.

Note: these constants previously had no references anywhere. They are also duplicated verbatim in `TheSuperHackersConstants.cs` (identical content, different class name) — left alone here, worth deleting separately.

## Testing

- `GameSettingsMapperTests` 13/13 pass; full `GenHub.Tests.Core` 1541 passed / 2 failed, both failures (`NativeLaunchDiagnosticsTests`, `RetailArchiveRootTests`) reproducing unchanged on clean `development` as macOS environment failures.
- Test targets are seeded with sentinel values (inverted for booleans) so a missing or reverted mapper assignment fails rather than passing on the constructor default. Verified by mutation: restoring `?? false` on `CursorCaptureEnabledInFullscreenGame` fails both toggle tests; deleting the `SystemTimeFontSize` assignment fails all three numeric tests.

## Related

#359 backports this change to `release/alpha-4` **in full** — all three commits, diff byte-for-byte identical to this PR. That includes the behavioural defaults (observer, cursor capture, edge scroll), not just the font sizes.

Including them is deliberate: every value in the table above is verified against the engine source, and `release/alpha-4` carries the bug, so shipping alpha 4 without them means releasing a launcher that disables cursor capture and fullscreen edge scrolling on every Zero Hour launch. Note that `ci.yml` triggers only on `main` and `development`, so no build or test job runs against that base — see #359 for the local verification.

`MoneyTransactionVolume` still uses `?? 50` while the model declares `0`. The engine derives it from `AudioSettings` at runtime rather than a literal, so it needs someone familiar with the audio INI data to settle. Left untouched.

This PR aligns TheSuperHackers settings written at launch and displayed in the profile editor with the declared model and engine defaults.
- Replaces mapper fallback literals with centralized constants.
- Aligns font-size, observer, cursor-capture, and edge-scroll defaults across the mapper and view model.
- Adds tests covering unset defaults, model alignment, and explicit-value preservation.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| GenHub/GenHub.Core/Constants/GameSettingsTheSuperHackersConstants.cs | Adds named defaults for observer, cursor-capture, and edge-scroll settings consistent with the settings model. |
| GenHub/GenHub.Core/Helpers/GameSettingsMapper.cs | Replaces divergent literal fallbacks with the centralized TheSuperHackers defaults while preserving explicit profile values. |
| GenHub/GenHub.Tests/GenHub.Tests.Core/Helpers/GameSettingsMapperTests.cs | Adds coverage for fallback defaults, model-default consistency, and explicit font and toggle values. |
| GenHub/GenHub/Features/GameProfiles/ViewModels/GameSettingsViewModel.cs | Aligns initial editor values with the same centralized defaults used by launch-time mapping. |

</details>

<sub>Reviews (3): Last reviewed commit: ["fix(settings): align TheSuperHackers cur..."](https://github.com/community-outpost/genhub/commit/9f11b57421c261de03a6b6759d0e2398c15cbb34) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51175583)</sub>

**Context used:**

- Knowledge Base — [Game Profiles and Launching](https://app.greptile.com/genhub/-/custom-context/knowledge-base/community-outpost/genhub/-/docs/game-profiles-launching.md)

### DIFF
--- a/GenHub/GenHub.Core/Constants/GameSettingsTheSuperHackersConstants.cs
+++ b/GenHub/GenHub.Core/Constants/GameSettingsTheSuperHackersConstants.cs
@@ -44,4 +44,46 @@ public static class GameSettingsTheSuperHackersConstants
     /// Default font size for system time display.
     /// </summary>
     public const int DefaultSystemTimeFontSize = 8;
+
+    /// <summary>
+    /// Default for whether player observer mode is enabled.
+    /// Matches the engine fallback in OptionPreferences::getPlayerObserverEnabled.
+    /// </summary>
+    public const bool DefaultPlayerObserverEnabled = true;
+
+    /// <summary>
+    /// Default for cursor capture in fullscreen game.
+    /// Included in the engine's CursorCaptureMode_Default mask.
+    /// </summary>
+    public const bool DefaultCursorCaptureEnabledInFullscreenGame = true;
+
+    /// <summary>
+    /// Default for cursor capture in fullscreen menu.
+    /// Included in the engine's CursorCaptureMode_Default mask.
+    /// </summary>
+    public const bool DefaultCursorCaptureEnabledInFullscreenMenu = true;
+
+    /// <summary>
+    /// Default for cursor capture in windowed game.
+    /// Included in the engine's CursorCaptureMode_Default mask.
+    /// </summary>
+    public const bool DefaultCursorCaptureEnabledInWindowedGame = true;
+
+    /// <summary>
+    /// Default for cursor capture in windowed menu.
+    /// Absent from the engine's CursorCaptureMode_Default mask.
+    /// </summary>
+    public const bool DefaultCursorCaptureEnabledInWindowedMenu = false;
+
+    /// <summary>
+    /// Default for screen edge scrolling in a fullscreen app.
+    /// The engine's ScreenEdgeScrollMode_Default is exactly this flag.
+    /// </summary>
+    public const bool DefaultScreenEdgeScrollEnabledInFullscreenApp = true;
+
+    /// <summary>
+    /// Default for screen edge scrolling in a windowed app.
+    /// Absent from the engine's ScreenEdgeScrollMode_Default.
+    /// </summary>
+    public const bool DefaultScreenEdgeScrollEnabledInWindowedApp = false;
 }
--- a/GenHub/GenHub.Core/Helpers/GameSettingsMapper.cs
+++ b/GenHub/GenHub.Core/Helpers/GameSettingsMapper.cs
@@ -209,17 +209,17 @@ public static class GameSettingsMapper
         settings.ArchiveReplays = profile.TshArchiveReplays ?? false;
         settings.MoneyTransactionVolume = profile.TshMoneyTransactionVolume ?? 50;
         settings.ShowMoneyPerMinute = profile.TshShowMoneyPerMinute ?? false;
-        settings.PlayerObserverEnabled = profile.TshPlayerObserverEnabled ?? false;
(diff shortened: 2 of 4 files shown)


bobtista:
Addressed both review points:

- Both default tests now seed the target with sentinel values (`99`) before mapping, so a removed or missing mapper assignment fails instead of silently passing on the constructor default. Verified by temporarily deleting the `SystemTimeFontSize` assignment: all three tests fail, where previously two of them passed.
- `TshResolutionFontAdjustment` added to the explicit-passthrough test.

PR #285, opened 2026-03-16
fix: reorganize settings UI, fix content priority resolution, and eliminate CAS warnings

undead2146:
## Overview

This PR resolves four GitHub issues related to settings UI organization, content priority , and CAS storage warnings.

### Settings UI Reorganization
- **#281**
Moved "Check for updates on startup" setting from Appearance section to Updates section where it logically belongs
- **#282**
Moved "Settings File Location" configuration from Appearance section to a new "Data Directories" section for better
organization

### Version Widget Enhancement
- **#283**:
Made the version widget in the bottom-right corner clickable
  - Clicking now copies the full version (including git hash) to clipboard
  - Shows a success notification when copied
  - Added tooltip "Click to copy version to clipboard"
  - Maintains original visual appearance with hand cursor on hover
 
### Content Priority Fix
- **#284**: 
Fixed mods not taking effect in-game due to incorrect file priority resolution
  - Root cause: `FullCopyStrategy` was sorting by enum value instead of content type priority
  - Changed to use `ContentTypePriority.GetPriority()` for proper ordering
  - Ensures mod files (priority 100) correctly override base game files (priority 10)
  - Priority order: Mod (100) > Patch (90) > GameClient (50) > Addon (40) > GameInstallation (10)
 
### CAS Storage Warning Fix
- Eliminated "Could not resolve source path for manifest" warnings for CAS-stored content
  - Added `source.path` marker file creation after CAS storage
  - Updated `ContentManifestPool` to handle CAS-ONLY marker gracefully
  - Prevents spurious warnings in logs for downloaded content
 
## Related Issues
Closes #281
Closes #282
Closes #283
Closes #284

<details open><summary><h3>Greptile Summary</h3></summary>

This PR addresses four GitHub issues: settings UI reorganization (#281, #282), a clickable version widget (#283), a content priority fix (#284), and CAS storage warning elimination. The core logic fixes are solid — `ContentTypePriority` is now exhaustively mapped and `FullCopyStrategy` correctly uses semantic priority ordering. However, two issues remain:

- **`"CAS-ONLY"` magic string**: The sentinel value written by `ContentStorageService` and compared in `ContentManifestPool` is hardcoded in both places rather than extracted to a shared constant in `FileTypes.cs` (which this same PR already uses for `SourcePathFileName`).
[...]

PR #382, opened 2026-08-18
fix(launching): only rewrite GeneralsOnline settings for GeneralsOnline profiles

bobtista:
## Problem

`ApplyGeneralsOnlineSettingsAsync` was gated only on `GameType == ZeroHour`, so launching **any** Zero Hour profile — retail, TheSuperHackers, CommunityOutpost — rewrote the GeneralsOnline client's `settings.json`.

That file is a single global file owned by the GeneralsOnline client, not a per-profile file. The launcher also built a **fresh** `GeneralsOnlineSettings` and saved it wholesale (`SaveGeneralsOnlineSettingsAsync` does a plain serialize + write, no read-modify-write), and the mapper coalesced every unset profile field to a hardcoded default. Net effect: every Zero Hour launch reset whatever the user had configured inside the GeneralsOnline client.

The settings screen had the same defect on its own path, and two ways to write wrong values, both fixed here.

## Changes

**Gating**
- Gate on the profile actually being a GeneralsOnline profile. Rather than inlining a `PublisherType` comparison, this promotes the existing `IsGeneralsOnlineProfile` predicate out of `ProfileLauncherFacade` into `GameProfileExtensions` and reuses it — it is a three-tier check (publisher type, then client name, then enabled content) whose fallbacks exist for legacy/incomplete profiles, so a bare `PublisherType` check would have silently stopped applying settings for those.
- **The publisher type is now authoritative.** The fallbacks previously applied even when `PublisherType` explicitly named a *different* publisher, so a TheSuperHackers or CommunityOutpost profile that merely had GeneralsOnline content enabled was classified as GeneralsOnline — recreating the same cross-client data loss. The name/content heuristics now apply only when the publisher type is absent. The `ProfileLauncherFacade` call site is unaffected: it only reaches the predicate inside the `else` of a blank-publisher check.
- **`GameSettingsViewModel.SaveSettings` is gated too.** It previously loaded and rewrote the global `settings.json` on every save regardless of game type or publisher — the same bug on the settings-screen path.

**Preserving what the user configured**
- Load existing settings and merge, falling back to a new instance only on failure, mirroring the Options.ini path in the same file.
- The five `ApplyGo*` mapper methods now use per-field `HasValue` guards, matching the idiom `ApplyTshToOptions` already uses, so a merge is not immediately flattened.
- `[JsonExtensionData]` on `GeneralsOnlineSettings` **and its five nested classes**, so client-owned keys survive the round-trip (root-only would still have dropped keys inside those objects).
- `GameSettingsViewModel` captures and mutates the loaded settings rather than building a new object, and reads `settings.json` first when initialised from a profile and never loaded.
- **View model defaults no longer fabricate values.** `_goShowPing`, `_goShowPlayerRanks`, `_goRememberUsername`, `_goEnableNotifications` and `_goEnableSoundNotifications` defaulted to `false` while the model defaults them to `true`, so an unconditional save could flip a user's real toggles off. Both the model and the view model now reference shared constants so they cannot drift, and `InitializeForProfileAsync` seeds the view model from the client's own `settings.json` before overlaying the profile's declared values. That matters in both directions — aligning defaults alone stops `true` becoming `false`, but only the seeding stops a deliberate `false` becoming `true`.
- **`MoneyTransactionVolume` default made explicit.** Removing the mapper's coalesce would have silently changed the effective default for a fresh `settings.json` from 50 to the model's 0. The evidence points to 50 being intended (the old mapper, the view model field, and a 0–100 slider where 0 means muted), so 50 is now declared on the model via a constant and shared, rather than living as a literal in one code path.

Nullable-ing the view model's 25 bound bool/int properties to get true `HasValue` semantics was considered and rejected for a release fix: it is a large change to the binding surface and would not actually disambiguate "unticked" from "never touched". Seeding from the client's real settings achieves the same guarantee without that risk.

## Tests

1638 -> 1660, all passing (including the real-engine launch tests). Every guard was verified to bite by reverting it and confirming exactly the intended tests fail.

- non-GeneralsOnline Zero Hour profile writes no GeneralsOnline settings; a GeneralsOnline profile does
- an explicit non-GeneralsOnline publisher is **not** reclassified by a GeneralsOnline client name or enabled content
- saving from the settings screen for TheSuperHackers/CommunityOutpost (Zero Hour and Generals) leaves `settings.json` untouched
- values the profile does not specify survive a launch and a save; enabled toggles cannot be flipped off
- unknown root and nested JSON keys survive a load/modify/save round-trip
- unset money transaction volume stays audible

## Notes

`TheSuperHackersSettings` has the same shape of gap but not the same fix — it round-trips through the `[TheSuperHackers]` section of Options.ini rather than JSON, and both writers assign a freshly built dictionary over that section, so unmodelled keys there are dropped. That needs a dictionary merge and is left out of scope.

Rebased onto `development` after #285, which refactored `ProfileLauncherFacade`; the predicate call site was re-applied against the new structure.

### DIFF
--- a/GenHub/GenHub.Core/Constants/GameSettingsGeneralsOnlineConstants.cs
+++ b/GenHub/GenHub.Core/Constants/GameSettingsGeneralsOnlineConstants.cs
@@ -16,6 +16,18 @@ public static class GameSettingsGeneralsOnlineConstants
     public const string TemporarySettingsFileExtension = ".tmp";
 
     /// <summary>
+    /// Number of times the completed settings file is moved over settings.json before the save
+    /// gives up and reports the failure. Anything holding settings.json open releases it within
+    /// milliseconds, so a handful of attempts either succeeds or is looking at a real fault.
+    /// </summary>
+    public const int SettingsReplaceAttemptLimit = 5;
+
+    /// <summary>
+    /// Delay between attempts to move the completed settings file over settings.json.
+    /// </summary>
+    public const int SettingsReplaceRetryDelayMilliseconds = 20;
+
+    /// <summary>
     /// Default chat font size.
     /// </summary>
     public const int DefaultChatFontSize = 12;
--- a/GenHub/GenHub.Core/Constants/GameSettingsGeneralsOnlineConstants.cs
+++ b/GenHub/GenHub.Core/Constants/GameSettingsGeneralsOnlineConstants.cs
@@ -11,6 +11,11 @@ public static class GameSettingsGeneralsOnlineConstants
     public const string SettingsFileName = "settings.json";
 
     /// <summary>
+    /// Extension of the file a save is written to before it is moved over settings.json.
+    /// </summary>
+    public const string TemporarySettingsFileExtension = ".tmp";
+
+    /// <summary>
     /// Default chat font size.
     /// </summary>
     public const int DefaultChatFontSize = 12;
--- a/GenHub/GenHub.Core/Constants/GameSettingsGeneralsOnlineConstants.cs
+++ b/GenHub/GenHub.Core/Constants/GameSettingsGeneralsOnlineConstants.cs
@@ -24,4 +24,29 @@ public static class GameSettingsGeneralsOnlineConstants
     /// Maximum chat font size.
     /// </summary>
     public const int MaxChatFontSize = 24;
+
+    /// <summary>
+    /// Default for whether ping is shown.
+    /// </summary>
+    public const bool DefaultShowPing = true;
+
+    /// <summary>
+    /// Default for whether player ranks are shown.
+    /// </summary>
+    public const bool DefaultShowPlayerRanks = true;
+
+    /// <summary>
+    /// Default for whether the username is remembered.
+    /// </summary>
+    public const bool DefaultRememberUsername = true;
+
+    /// <summary>
+    /// Default for whether notifications are enabled.
+    /// </summary>
+    public const bool DefaultEnableNotifications = true;
+
+    /// <summary>
+    /// Default for whether sound notifications are enabled.
+    /// </summary>
+    public const bool DefaultEnableSoundNotifications = true;
 }
--- a/GenHub/GenHub.Core/Constants/GameSettingsTheSuperHackersConstants.cs
+++ b/GenHub/GenHub.Core/Constants/GameSettingsTheSuperHackersConstants.cs
@@ -46,6 +46,12 @@ public static class GameSettingsTheSuperHackersConstants
     public const int DefaultSystemTimeFontSize = 8;
 
     /// <summary>
+    /// Default volume for money transaction audio events, on the same 0-100 scale the settings
+    /// screen exposes. Zero would mute them, which is a choice rather than a default.
+    /// </summary>
+    public const int DefaultMoneyTransactionVolume = 50;
+
+    /// <summary>
     /// Default for whether player observer mode is enabled.
     /// Matches the engine fallback in OptionPreferences::getPlayerObserverEnabled.
     /// </summary>
--- a/GenHub/GenHub.Core/Extensions/GameProfileExtensions.cs
+++ b/GenHub/GenHub.Core/Extensions/GameProfileExtensions.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Linq;
+using GenHub.Core.Constants;
 using GenHub.Core.Models.GameProfile;
 
 namespace GenHub.Core.Extensions;
@@ -19,6 +22,32 @@ public static class GameProfileExtensions
                HasCustomTshSettings(profile) ||
                HasCustomGeneralsOnlineSettings(profile) ||
                HasCustomNetworkSettings(profile);
+    }
+
+    /// <summary>
+    /// Checks if a profile runs the GeneralsOnline client.
+    /// </summary>
+    /// <remarks>
+    /// The publisher type is the authoritative marker, but profiles created before it was
+    /// recorded carry it only in the client name or the enabled content ids, so both are
+    /// consulted as fallbacks.
+    /// </remarks>
+    /// <param name="profile">The game profile.</param>
+    /// <returns>True if the profile runs GeneralsOnline, false otherwise.</returns>
+    public static bool IsGeneralsOnlineProfile(this GameProfile profile)
+    {
+        if (string.Equals(profile.GameClient?.PublisherType, PublisherTypeConstants.GeneralsOnline, StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
+        if (profile.GameClient?.Name?.Contains(PublisherTypeConstants.GeneralsOnline, StringComparison.OrdinalIgnoreCase) == true)
+        {
+            return true;
+        }
+
+        return profile.EnabledContentIds?
+            .Any(id => id.Contains(PublisherTypeConstants.GeneralsOnline, StringComparison.OrdinalIgnoreCase)) == true;
     }
 
     private static bool HasCustomVideoSettings(GameProfile profile)
--- a/GenHub/GenHub.Core/Extensions/GameProfileExtensions.cs
+++ b/GenHub/GenHub.Core/Extensions/GameProfileExtensions.cs
@@ -28,17 +28,20 @@ public static class GameProfileExtensions
     /// Checks if a profile runs the GeneralsOnline client.
     /// </summary>
     /// <remarks>
-    /// The publisher type is the authoritative marker, but profiles created before it was
-    /// recorded carry it only in the client name or the enabled content ids, so both are
-    /// consulted as fallbacks.
+    /// A recorded publisher type settles the question either way. The client name and the enabled
+    /// content ids are consulted only when no publisher type was recorded, which is the case for
+    /// profiles created before it existed: a TheSuperHackers profile with GeneralsOnline content
+    /// enabled belongs to TheSuperHackers, and answering otherwise would let it rewrite the
+    /// GeneralsOnline client's global settings.
     /// </remarks>
     /// <param name="profile">The game profile.</param>
     /// <returns>True if the profile runs GeneralsOnline, false otherwise.</returns>
     public static bool IsGeneralsOnlineProfile(this GameProfile profile)
     {
-        if (string.Equals(profile.GameClient?.PublisherType, PublisherTypeConstants.GeneralsOnline, StringComparison.OrdinalIgnoreCase))
+        var publisherType = profile.GameClient?.PublisherType;
+        if (!string.IsNullOrWhiteSpace(publisherType))
         {
-            return true;
+            return string.Equals(publisherType, PublisherTypeConstants.GeneralsOnline, StringComparison.OrdinalIgnoreCase);
         }
 
         if (profile.GameClient?.Name?.Contains(PublisherTypeConstants.GeneralsOnline, StringComparison.OrdinalIgnoreCase) == true)
--- a/GenHub/GenHub.Core/Helpers/GameSettingsMapper.cs
+++ b/GenHub/GenHub.Core/Helpers/GameSettingsMapper.cs
@@ -89,6 +89,11 @@ public static class GameSettingsMapper
     /// Applies settings from a GameProfile to a GeneralsOnlineSettings object.
     /// Used by GameLauncher to prepare settings.json for launch.
     /// </summary>
+    /// <remarks>
+    /// Only the fields the profile declares are written, as <see cref="ApplyToOptions"/> does for
+    /// Options.ini. The caller passes the settings already on disk, and anything the profile leaves
+    /// unset is the GeneralsOnline client's own configuration, which a launch must not overwrite.
+    /// </remarks>
     /// <param name="profile">The GameProfile source.</param>
     /// <param name="settings">The GeneralsOnlineSettings to populate.</param>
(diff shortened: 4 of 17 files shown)


bobtista:
Triaged the bot findings on this PR. Four were real and are now fixed; the rest were either already addressed by later commits on the branch or don't hold up.

### Fixed

- **A failed `settings.json` read still fell through to a wholesale rewrite** (`a08ec9d0`). `LoadGeneralsOnlineSettingsAsync` reports success for a missing file, so a *failure* means the client's own file exists and could not be read — locked, permissions, malformed. The old fallback then rewrote it from defaults, discarding exactly the client-owned keys this PR exists to preserve. Both the launcher and the settings-screen save path now abort the write and surface the error. This one was raised against the first commit but genuinely survived the later gating commit, which only fixed the *unconditional* write.
- **Null nested sections** (`396212b2`). A `settings.json` containing `"camera": null` deserializes over the `= new()` initializer, so mutating `settings.Camera.*` threw and the whole save failed. Sections are now re-initialized at the deserialization boundary and on both merge paths.
- **Docs** (`3b3cb0a3`) — the troubleshooting entry in `game-settings-architecture.md` documented the opposite semantics (`HasValue` as the bug, `?? default` as the fix) and now describes load-then-merge.
- **Test quality** (`3f4eb790`) — assert key presence before indexing `AdditionalSettings`, include `FirstError` in bare success assertions, and cover a `null` publisher type rather than only the empty string.

### Not reproducible / already addressed

- **`IsGeneralsOnlineProfile` lets fallbacks override an explicit publisher (High).** The current implementation short-circuits before reaching them:
  ```csharp
  var publisherType = profile.GameClient?.PublisherType;
  if (!string.IsNullOrWhiteSpace(publisherType))
  {
      return string.Equals(publisherType, PublisherTypeConstants.GeneralsOnline, StringComparison.OrdinalIgnoreCase);
  }
  ```
  A recorded publisher type settles it either way, so a TheSuperHackers profile with GeneralsOnline content enabled returns false. Pinned by `IsGeneralsOnlineProfile_WithOtherPublisherAndGeneralsOnlineHints_ReturnsFalse`, which uses a TheSuperHackers profile named "GeneralsOnline Compatible" with a `generalsonline` content id. Fixed in `45b01407`.
- **`MoneyTransactionVolume` defaults to 0 (muted).** Already resolved in the commit this was posted against: the model now defaults to `GameSettingsTheSuperHackersConstants.DefaultMoneyTransactionVolume = 50`, and the view model seeds from the same constant, so the launcher and the settings screen agree.
- **View model writes GeneralsOnline settings unconditionally.** Gated in `f07de08f` on `SelectedGameType == ZeroHour && _currentProfileIsGeneralsOnline`, covered by `SaveSettings_Should_NotWriteGeneralsOnlineSettings_ForOtherPublishersAsync`.
- **View model boolean defaults mismatched the model.** Fixed in `20dd35f5`; all five now seed from `GameSettingsGeneralsOnlineConstants`.

### Declined

**Case-variant extension keys producing duplicate JSON properties.** Extension data only ever receives keys that failed to match a modeled property, and they are written back verbatim, so a load/save round trip cannot emit an exact duplicate. Producing a variant would require the client to write the same setting twice in different casings, or to switch its whole file away from snake_case — in which case every modeled setting is already unmapped and this is the lesser problem. Enabling `PropertyNameCaseInsensitive` is the riskier change, since it silently reclassifies which keys count as modeled. Preservation is unaffected either way.

Suite is at 1665 passing (+8 new cases); the only failures are the real-engine launch tests, which are environment-dependent and unrelated.

bobtista:
Second round addressed. The cache findings were right, and rather than patch the three symptoms I removed the cache.

### The cache is gone

`_currentGeneralsOnlineSettings` was conflating two jobs, which is why it produced three separate findings. Split them:

- **The rewrite baseline** is now re-read from disk on every save. A snapshot of a global file that the GeneralsOnline client itself writes can only ever be equal to or staler than a fresh read, and the read is one file open on a click-driven path. That removes both the stale-rewrite and previous-profile-snapshot cases with no invalidation rules to get wrong.
- **A seeding flag** covers what re-reading alone does not. The view model has no unset state, so it writes all 24 modeled fields; if the seeding read failed, a successful save-time read would still have had the view model's defaults stamped over the user's real values. An unseeded save now aborts the rewrite and says why in `StatusMessage`.

The redundant reassignment fell out naturally, and `CreateGeneralsOnlineSettings` became `MergeViewModelIntoGeneralsOnlineSettings`, which is honest about mutating the just-read instance.

Two existing tests were strengthened because they lacked teeth: the read-before-rewrite test used `Times.AtLeastOnce` and would have passed with the save-time read deleted, and the cannot-be-read test was exercising the new unseeded gate rather than the path in its name.

### Atomic write yes, lock no

`File.WriteAllTextAsync` truncates first, so an interruption leaves the client with an unreadable settings.json including keys GenHub cannot reconstruct. Now serialized to a uniquely-named temp beside the file and moved over it, temp cleaned in `finally`. On Windows this also removes a real failure today: `WriteAllTextAsync` opens with `FileShare.Read`, so a second concurrent writer currently gets an `IOException`.

I declined the lock deliberately. What it prevents is last-writer-wins on the modeled fields when two launches race — but both writers loaded recently so both preserve the client's unknown keys, and settings.json is a single global file that can only hold one configuration; the profile that launched last winning is arguably correct. Moving load-merge-save behind a shared semaphore would also make the mocked Load/Save verifications in six existing tests unobservable, trading real coverage for a benign race. The codebase precedent agrees: Options.ini takes its semaphore per-operation and does not protect its own read-modify-write either.

**Caveat on the new atomicity tests:** I checked them against a reverted implementation and both still pass on macOS — .NET does not enforce the share mode there and the payload is too small to tear. They have teeth on Windows CI, where the sharing violation is real, but they do not prove atomicity on a Mac. Flagging that rather than claiming coverage I do not have.

### Rejected — case-variant extension keys

Verified empirically rather than by inspection. With `PropertyNamingPolicy = SnakeCaseLower`, deserializing `{"ShowPing": true}`:

```
caseInsensitive=False  extraKeys=[ShowPing]  roundTrip={"show_ping":false,"ShowPing":true}
caseInsensitive=True   extraKeys=[ShowPing]  roundTrip={"show_ping":false,"ShowPing":true}
```

`PropertyNameCaseInsensitive` changes nothing, because the comparison target is the *converted* name `show_ping`, and `ShowPing` differs from it by an underscore rather than by case — so the suggested remedy is a no-op against its own motivating example. The round trip emits two textually distinct keys, not duplicates. The alternative remedy, dropping extension keys that case-insensitively match a modeled name, would silently delete a key the client owns — the exact data-loss class this PR exists to prevent.

Assert-before-index (`GameLauncherTests.cs:1051`, `GameSettingsViewModelTests.cs:429`, `GeneralsOnlineSettingsTests.cs:123-125`), the `null` publisher case (`[InlineData(null)]` at lines 39/78/95), the bare `Assert.True`, the test/doc mismatch, the five mismatched view-model defaults, `MoneyTransactionVolume`, the null nested sections NRE, and the publisher-authority remark.

### Also fixed while in there

The doc claimed `ApplyToOptions()` still coalesces; `ApplyTshToOptions` in fact uses `HasValue` guards throughout, same shape as the settings.json path. Corrected, and added a test for a profile with no `GameClient`, which the new-profile flow really does produce.

1672 passing; the only failures are the environment-dependent real-engine tests.

bobtista:
The Windows build caught a real regression from the atomic write in the previous round, and fixing it corrected part of that round's reasoning. Writing it up because the diagnosis is the interesting bit.

### The failure

`SaveGeneralsOnlineSettingsAsync_..._WhenSavesOverlapAsync` failed on the Windows runner only: 2 of 16 concurrent saves returned "Access to the path is denied". Two concurrent `File.Move(..., overwrite: true)` calls onto the same destination can be refused transiently on Windows. On macOS and Linux that call is `rename(2)` and cannot fail, which is why it passed everywhere else.

The atomic write did not *create* concurrency failure — the previous `File.WriteAllTextAsync` opens with `FileShare.Read`, so a second concurrent writer already got an `IOException`. It changed the shape of the failure. Reverting is not an option: a torn settings.json loses client-owned keys GenHub cannot reconstruct, which is the point of this PR.

### The part the previous round got wrong

That round declined a lock, reasoning that last-writer-wins is acceptable for a single global file. That holds for the *merge* semantics and still does — but it did not anticipate that the replacement itself can be refused, and it missed the read side entirely: `LoadGeneralsOnlineSettingsAsync` uses `File.ReadAllTextAsync`, which opens with `FileShare.Read` and therefore denies writers. A load overlapping a replace is an active *cause* of replace failures. A retry alone would have made the 16-way test pass by luck while leaving the mechanism in place.

### The fix — a narrow lock and a bounded retry

They cover different sources and neither alone delivers the contract.

**The lock** is the precedent already in this file: `_optionsIniWriteSemaphore` has serialized Options.ini reads and writes since before this PR. settings.json was simply the one global file without an equivalent, and that asymmetry is the actual defect. It guards the load and the save, and is released between them, so merge semantics are unchanged. It lives inside the concrete service, so the mocked `IGameSettingsService` Load/Save verifications in the GameLauncher and view-model tests never touch it — which was the previous round's stated objection to a lock, and it does not apply at this placement.

**The retry** covers what no in-process lock can: settings.json belongs to the GeneralsOnline client, and a running client, a virus scanner or the search indexer holding it open makes Windows refuse the replace. Bounded at 5 attempts, 20 ms apart, catching only `IOException`/`UnauthorizedAccessException`, with the final attempt outside the loop so a genuine fault propagates with its real exception instead of a synthesized one.

Also hardened the temp-file cleanup: it ran in a `finally` and caught only `IOException`, so an `UnauthorizedAccessException` during cleanup would have masked the real error being reported.

### On the test's contract

The test asserted that all 16 overlapping saves succeed while being named "should leave readable settings" — a stronger claim than the name. I concluded the strong assertion is the correct contract, since a launch that reports a settings failure has silently lost the settings the user chose for that profile, so the **name** was corrected to `..._Should_SucceedForEverySave_WhenSavesOverlapAsync` rather than the assertion weakened to match. Which save wins remains undefined; none may be turned away.

Two tests added: 16 saves and 16 loads concurrently against a seeded file (the regression guard for the read/write half), and a replacement that can never succeed — a directory placed at the settings.json path — pinning that retries are bounded, the failure surfaces rather than being swallowed, and no temp file is left behind.

**Honest scope:** the two overlap tests pass trivially on macOS and Linux and only *prove* the fix on Windows CI. The bounded-retry test is genuinely cross-platform.

### Options.ini — related, deliberately not changed here

`SaveTheSuperHackersSettingsAsync` has the same non-atomic load-merge-save shape, but it cannot hit this failure: it never calls `File.Move` and its access is already serialized. Its distinct weakness is the opposite one — `SaveOptionsAsync` uses the truncating `File.WriteAllLinesAsync`, exactly what settings.json was moved away from, and it round-trips `AdditionalSections` so an interrupted write loses user-owned keys. The game itself also writes Options.ini on exit, which an in-process semaphore does nothing about. Converting it would widen this PR onto an unrelated risk I cannot test from here, so it is worth a follow-up issue rather than a change now.

1674 passing; the only failures are the environment-dependent real-engine tests.

undead2146 (dismissed):
LGTM

bobtista:
Fixed, and the same shape turned out to exist in the reverse direction in the same method.

### Reporting the partial outcome, not aborting early

I rejected the "check the gate before writing Options.ini" option. Reordering does not make the operation atomic — `SaveGeneralsOnlineSettingsAsync` can still fail *after* Options.ini is written, so it converts one specific failure into a total abort and leaves the other partial window open. Two files, two writes, no transaction: the misreporting has to be fixed at the reporting layer regardless. It also costs the user real capability — a locked `settings.json` would permanently block saving unrelated video and audio changes for a profile whose GeneralsOnline half they may not care about.

### The reverse partial was already live

The GeneralsOnline read-merge-write runs unconditionally **after** `SaveOptionsAsync`, including when that failed. So `settings.json` could be rewritten successfully while Options.ini was refused, and the status still said `"Failed to save settings: <options error>"` — again claiming nothing landed. Same misreporting, opposite direction, not mentioned in the finding. Fixed in the same pass.

`SaveSettings` now classifies the halves independently and reports four distinct outcomes: both saved, Options.ini only, settings.json only, or neither. Non-GeneralsOnline profiles are untouched.

One correctness fix rode along: `_currentOptions` and `OptionsFileExists` were only updated on *total* success, so after a partial save the view model's baseline for the next save was stale relative to the file that had actually been written. They now follow the Options.ini write.

Added a `<remarks>` recording why the writes are reported separately rather than reordered, so the next reader does not "fix" it back into an early abort.

### Tests

Four new, covering each combination, plus a guard that the split reporting does not soften a genuine total failure. Verified non-vacuous by reverse-applying the patch: the three behavioural tests and one updated existing test all fail without the fix.

Worth noting one of those updates — `SaveSettings_Should_NotRewriteGeneralsOnlineSettings_WhenSeedingFailedAsync` was asserting the misleading `"Failed to save settings"` string. The test was encoding the bug.

1681 passing.

### Adjacent gap, deliberately not fixed

`LoadSettings` has the same honesty problem on the read side: when Options.ini loads but `LoadGeneralsOnlineSettingsAsync` fails, the failure is only logged and the status still reads `"Loaded ZeroHour settings from Options.ini"`. The user sees the GeneralsOnline tab populated with view-model defaults, has no indication they are not the client's real values, and only discovers it when a later save refuses to write them. Surfacing the seeding failure at load time is what would let them act — close the client, unlock the file — before they start editing. Out of scope here, worth a follow-up.

undead2146 (approved):
LGTM

PR #383, opened 2026-08-18
fix(userdata): copy CAS content to user-writable targets and guard destructive data deletion

bobtista:
## Problem

Three defects around user data, the first of which silently corrupts content-addressable storage.

**1. Mutable user data was hard-linked directly to CAS objects.** `UserDataTrackerService` materialized content with `useHardLink: true` into the user's real game data directory (`Documents\Command and Conquer Generals Zero Hour Data`) — the directory the game engine and GenHub both write to. A hard link shares the inode, and `GameSettingsService` writes there with `File.WriteAllLinesAsync`, which truncates in place. Any such write therefore rewrote the canonical CAS object, breaking the hash-to-content invariant for **every profile** referencing that hash, with no detection and no user-visible symptom until hashes started failing. GeneralsOnline routes its data patch (community balance and core INI files) through exactly this path.

**2. "Delete All Application Data" had no confirmation.** The button was bound straight to the command — one click, irreversible, and it discards the pristine game data backups.

**3. A hash mismatch silently forfeited the user's backup.** In `CleanupInstalledFilesAsync`, if the deployed file failed hash verification (i.e. the user or game had modified it) the file was left in place and only logged; the restore was then guarded by `!File.Exists(...)`, which was consequently false, so the backup was never restored — and `DeleteAllUserDataAsync` then deleted the backups directory. The modified file stayed and the pristine original was gone forever. Defect 1 is what makes those files get modified in the first place, so the two compound.

## Changes

- **Copy, never link, for user-writable destinations.** New `ContentInstallTargetExtensions.IsUserWritableTarget()` covers `UserDataDirectory`, `UserMapsDirectory`, `UserReplaysDirectory` and `UserScreenshotsDirectory`; both materialization paths funnel through a single helper that calls `CopyFromCasAsync` for those targets and never reaches `LinkFromCasAsync`. The decision is gated on the install target, not hardcoded.
- **Fixed the same defect class in the restore paths.** `RestoreBackupQuietly`, `RollbackActivatedFiles` and the deactivation restore used `File.Copy(..., overwrite: true)`, which truncates in place — on an existing hard-linked deployment that would write backup content *into* the CAS object on the first uninstall after upgrading. They now unlink the destination first. `FileOperationsService.CopyFileAsync` had the same hazard and now unlinks via `DeleteFileIfExists` rather than only handling reparse points.
- **Confirmation before deletion**, reusing the existing `IDialogService.ShowConfirmationAsync` pattern already used by profile deletion. The prompt states the action is irreversible and that backups are discarded, and deliberately has no `sessionKey` so it cannot be suppressed. The gate lives in the command, not the view, so it is unit-testable.
- **Hash mismatch now preserves both files.** An explicit restore-needed flag replaces the `!File.Exists(...)` guard; the modified file is moved aside to `<name>.user-modified` (collision-suffixed) and the backup is restored. `DeleteAllUserDataAsync` tracks failed restores and missing manifests, and on failure keeps the backups directory, deleting only the manifests and index, with a warning naming the retained path.

## Tests

1638 -> 1645, all passing.

The independent-copy test uses a real CAS directory and a link mock that creates a genuine hard link, so the copy-vs-link distinction is real rather than simulated. Verified it bites: flipping `IsUserWritableTarget` back to `false` for `UserDataDirectory` fails the assertion with the CAS object reading back `engine-rewrote-this-file...` instead of `pristine-cas-content`.

- user-writable target deploys an independent copy
- hash mismatch preserves the modified file and restores the backup
- delete-all retains backups when a restore failed, clears them when all succeeded
- delete-all deletes nothing when confirmation is declined, proceeds when accepted, and warns that backups are discarded

## Notes

An earlier revision of this branch also marked CAS objects read-only as defense in depth. That was removed deliberately: on NTFS a hard link shares one attribute record, and `HardLinkStrategy` links CAS objects straight into workspaces, so it would have made every workspace file read-only and changed workspace deletion behavior on Windows — untestable from this machine, and the guarantee decayed on the first workspace delete anyway. The copy-not-link change closes the reported corruption vector on its own. Worth revisiting separately with Windows coverage.

### DIFF
--- a/GenHub/GenHub.Core/Constants/AppConstants.cs
+++ b/GenHub/GenHub.Core/Constants/AppConstants.cs
@@ -133,6 +133,25 @@ public static class AppConstants
     public const string TokenFileName = ".ghtoken";
 
     /// <summary>
+    /// Title of the confirmation prompt shown before all application data is deleted.
+    /// </summary>
+    public const string DeleteAllDataConfirmationTitle = "Delete All Application Data";
+
+    /// <summary>
+    /// Body of the confirmation prompt shown before all application data is deleted.
+    /// </summary>
+    public const string DeleteAllDataConfirmationMessage =
+        "This permanently deletes every profile, workspace, manifest, CAS object and tracked user data " +
+        "installation. The pristine backups GenHub keeps of your original game data will be discarded " +
+        "as part of this, so anything GenHub replaced cannot be recovered afterwards.\n\n" +
+        "This action is irreversible. Continue?";
+
+    /// <summary>
+    /// Confirm button text for the delete-all-application-data prompt.
+    /// </summary>
+    public const string DeleteAllDataConfirmText = "Delete Everything";
+
+    /// <summary>
     /// Gets assembly metadata by key.
     /// </summary>
     private static string? GetAssemblyMetadata(string key)
--- a/GenHub/GenHub.Core/Constants/UserDataConstants.cs
+++ b/GenHub/GenHub.Core/Constants/UserDataConstants.cs
@@ -1,0 +1,13 @@
+namespace GenHub.Core.Constants;
+
+/// <summary>
+/// Constants for tracked user data installations.
+/// </summary>
+public static class UserDataConstants
+{
+    /// <summary>
+    /// Suffix appended to a deployed file that no longer matches its recorded hash when it is
+    /// moved aside so the pristine backup can be restored over it.
+    /// </summary>
+    public const string UserModifiedSuffix = ".user-modified";
+}
--- a/GenHub/GenHub.Core/Extensions/Enums/ContentInstallTargetExtensions.cs
+++ b/GenHub/GenHub.Core/Extensions/Enums/ContentInstallTargetExtensions.cs
@@ -11,15 +11,19 @@ public static class ContentInstallTargetExtensions
     /// Determines whether the target resolves to a directory the user and the game engine
     /// write to directly, which means deployed content must never share storage with the
     /// content-addressable object it originated from.
+    /// <para>
+    /// Only the two targets that are definitively not user data are listed as such: every other
+    /// value, including any added later, is treated as user-writable and therefore copied. That
+    /// matches the resolver, whose own default arm places unmapped targets inside the user data
+    /// root, and it fails towards an extra copy rather than towards a hard link into Documents.
+    /// </para>
     /// </summary>
     /// <param name="installTarget">The install target to inspect.</param>
     /// <returns><c>true</c> when the destination is user-writable; otherwise, <c>false</c>.</returns>
     public static bool IsUserWritableTarget(this ContentInstallTarget installTarget) => installTarget switch
     {
-        ContentInstallTarget.UserDataDirectory => true,
-        ContentInstallTarget.UserMapsDirectory => true,
-        ContentInstallTarget.UserReplaysDirectory => true,
-        ContentInstallTarget.UserScreenshotsDirectory => true,
-        _ => false,
+        ContentInstallTarget.Workspace => false,
+        ContentInstallTarget.System => false,
+        _ => true,
     };
 }
--- a/GenHub/GenHub.Core/Extensions/Enums/ContentInstallTargetExtensions.cs
+++ b/GenHub/GenHub.Core/Extensions/Enums/ContentInstallTargetExtensions.cs
@@ -1,0 +1,25 @@
+using GenHub.Core.Models.Enums;
+
+namespace GenHub.Core.Extensions.Enums;
+
+/// <summary>
+/// Provides extension methods for the <see cref="ContentInstallTarget"/> enum.
+/// </summary>
+public static class ContentInstallTargetExtensions
+{
+    /// <summary>
+    /// Determines whether the target resolves to a directory the user and the game engine
+    /// write to directly, which means deployed content must never share storage with the
+    /// content-addressable object it originated from.
+    /// </summary>
+    /// <param name="installTarget">The install target to inspect.</param>
+    /// <returns><c>true</c> when the destination is user-writable; otherwise, <c>false</c>.</returns>
+    public static bool IsUserWritableTarget(this ContentInstallTarget installTarget) => installTarget switch
+    {
+        ContentInstallTarget.UserDataDirectory => true,
+        ContentInstallTarget.UserMapsDirectory => true,
+        ContentInstallTarget.UserReplaysDirectory => true,
+        ContentInstallTarget.UserScreenshotsDirectory => true,
+        _ => false,
+    };
+}
--- a/GenHub/GenHub.Core/Interfaces/Workspace/IFileOperationsService.cs
+++ b/GenHub/GenHub.Core/Interfaces/Workspace/IFileOperationsService.cs
@@ -59,6 +59,20 @@ public interface IFileOperationsService
         CancellationToken cancellationToken = default);
 
     /// <summary>
+    /// Compares a file against an expected hash, distinguishing a genuine mismatch from a failure to
+    /// compute the hash at all. Callers that act destructively on a mismatch must use this rather
+    /// than <see cref="VerifyFileHashAsync"/>, which collapses both outcomes into <c>false</c>.
+    /// </summary>
+    /// <param name="filePath">The file path.</param>
+    /// <param name="expectedHash">The expected hash value.</param>
+    /// <param name="cancellationToken">A cancellation token.</param>
+    /// <returns>The verification outcome.</returns>
+    Task<FileHashVerification> CheckFileHashAsync(
+        string filePath,
+        string expectedHash,
+        CancellationToken cancellationToken = default);
+
+    /// <summary>
     /// Applies a patch to a target file. The patch format is determined by the implementation.
     /// </summary>
     /// <param name="targetPath">The file to be patched.</param>
--- a/GenHub/GenHub.Core/Interfaces/Workspace/IFileOperationsService.cs
+++ b/GenHub/GenHub.Core/Interfaces/Workspace/IFileOperationsService.cs
@@ -62,6 +62,8 @@ public interface IFileOperationsService
     /// Compares a file against an expected hash, distinguishing a genuine mismatch from a failure to
     /// compute the hash at all. Callers that act destructively on a mismatch must use this rather
     /// than <see cref="VerifyFileHashAsync"/>, which collapses both outcomes into <c>false</c>.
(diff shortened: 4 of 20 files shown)


bobtista:
Triaged all the bot findings. Every one held up, and they're fixed across 13 focused commits. Two of them were genuine regressions this branch introduced, so thanks for the catch.

### The two regressions

- **`CopyFileAsync` destroyed a valid destination on source failure** (`fe5268ec`). The unlink was moved above the source open, so a missing or unreadable source deleted the existing destination before the method threw, and a self-copy deleted the source before it could be read. The source stream is now opened first and a same-path guard no-ops the self-copy. The unlink itself is kept — it is load-bearing for breaking hard links to CAS.
- **A transient hash-verification failure was treated as "user modified"** (`b71b8617`). `VerifyFileHashAsync` returns `false` on *any* internal exception, including a file briefly locked by the running game, so the new branch could rename a pristine file to `.user-modified`, restore the backup over it, and log that the user's edits were preserved — none of which was true. Added a tri-state `CheckFileHashAsync` (`Failed` / `Match` / `Mismatch`); a file is only moved aside on a confirmed mismatch, and a verification failure now leaves both the file and its backup untouched.

### Data-safety fixes

- **Volume-unsafe restore** (`f5a98a21`) — the last `File.Move` restore path now goes through the same unlink-then-copy as its siblings. It matters because backups live under app data while the target is in Documents, which is frequently redirected to another drive or OneDrive, and in the hash-match branch the deployed file is already deleted by that point.
- **Cancellation during delete-all** (`dad955d5`) — cancellation now aborts before any manifests or index are removed, rather than being logged and continuing into the deletion branch while reporting success.
- **Uninstall discarded the cleanup signal** (`5ef15ba4`) — it now fails and keeps tracking data when a pristine backup cannot be restored. The signal itself was also weaker than it looked: the old `hasBackup && File.Exists(backupPath)` post-check reported success when a restore threw *after* the backup had already vanished, so it is now an explicit flag.
- **Stale index key blocked cleanup forever** (`30d68f3b`) — a missing manifest file counts as "nothing to restore"; only a manifest that exists and cannot be read blocks deletion.
- **Retained backups reported as success** (`f6d01c06`) — returns a failure naming the backups path, and the UI shows a partial-deletion error instead of "deleted successfully".
- **Rollback and deactivation left the backup and `BackupPath` recorded** (`087bfcf5`) — both now consume the backup, so a later uninstall no longer sees the restored original as a mismatch and duplicates it.
- **`IsUserWritableTarget` failed open** (`f473fb19`) — an unmapped target is now copied rather than hard-linked into Documents.
- Plus `DeleteAllData` try/catch (`fee0d209`), install-failure cleanup logging that the retained backups are referenced by no manifest (`daab4ec2`), and the `constants.md` entries (`8d17f028`).

### Where I differed on approach

- The suggested fix for the copy bug was stage-into-temp-then-atomically-replace. I used the reorder plus same-path guard instead: it closes the data-loss window without changing the atomicity contract callers already rely on, whereas staging would silently move which errors surface where.
- For the hash check, "let verification errors throw" would have changed `VerifyFileHashAsync` for six workspace and validation call sites that all depend on false-on-error. Adding the tri-state alongside leaves those untouched.

### Already addressed

The DeepSource "generic `Exception` caught" on `UserDataTrackerServiceSafetyTests.cs:323` was filed against `f58b7389` and resolved in `38640fe9`, which narrowed that helper to `IOException or UnauthorizedAccessException or EntryPointNotFoundException or DllNotFoundException`.

### Tests

15 added. The one worth calling out pins the dangerous window directly: the backup is removed *between* the move-aside and the restore, asserting the uninstall reports failure and keeps tracking data rather than silently leaving the original path empty. The decline/accept confirmation tests now assert all six deletion channels rather than three.

1657 passing; the only failures are the real-engine launch tests, which are environment-dependent here and unrelated to these changes.

bobtista:
Second round addressed. Seven fixes; the rest were already resolved earlier on this branch.

### The failure signal now reaches someone

The strongest finding was that the new failure result went nowhere. `CleanupProfileAsync` awaited `UninstallUserDataAsync` and dropped it, so "we could not restore your pristine backup, tracking data retained" never surfaced. It now aggregates and returns a failure naming the backups path, and `ProfileContentLinkerService` stops discarding it at both sites — the reinstall path aborts rather than redeploying over an unfinished restore, and the deselect path aggregates into its result. The chain terminates at `GameLauncher.TriggerBackgroundUserDataSwitch`, which logs the text, so the message reaches an operator instead of dying three levels down.

### Cancellation is no longer read as corruption

A cancelled `File.ReadAllTextAsync` inside the manifest load was caught by a broad handler and classified as "manifest unreadable", which drives the retention decision — cancellation is not corruption. The catch is now filtered so it propagates to the existing rethrow and aborts before manifests or the index are touched. Two adjacent holes fixed alongside: a cancelled read in `GetUserDataManifestAsync` previously became "no manifest, nothing to uninstall" and reported success, and `GetGameUserDataAsync` had the same shape.

### The retention path kept what its own message referenced

When backups are retained, the failure told the user to consult tracking data that the same branch then deleted. `manifests/` and `index.json` are now kept on that path, the cached index is only reset on the real delete path, and a retry after retention still clears everything — pinned by a new test so retention cannot become permanent.

### Smaller, all real

- A consumed backup whose delete fails no longer fails the restore that already succeeded; the stray file is logged instead.
- `CheckFileHashAsync` reports a missing file as `Failed`, not `Mismatch` — that distinction now decides whether a user's file is moved aside, so the contract matters. `VerifyFileHashAsync` is unchanged.
- `IsSamePath` resolves link targets before deciding a copy would overwrite its own source. The doc records the residual gap (hard links, 8.3 short names) and why opening the source before unlinking makes it fail safe. This also replaced the filtered broad catch DeepSource flagged with explicit clauses.
- The delete-all summary no longer claims success when user data was only partially deleted.

Cancellation during delete-all (`dad955d5` — the per-key OCE rethrow is ordered before the general catch, plus a method-level rethrow; the one genuine residual was the manifest-load hole fixed here), copy-failure-deletes-destination (`fe5268ec`), rollback stale backup (`087bfcf5`), uninstall discarding the signal (`5ef15ba4`), stale index key blocking cleanup (`30d68f3b`), transient verification treated as user modification (`b71b8617`), volume-unsafe `File.Move` (`f5a98a21`), missing try/catch on delete-all (`fee0d209`), fail-open install-target default (`f473fb19`), install-failure cleanup signal (`daab4ec2`), deactivation leaving the backup (`087bfcf5`), retained backups reported as success (`f6d01c06`), `constants.md` (`8d17f028`), and the 3-of-6 deletion channels (`70be4b87`).

On the request for a callback that removes the backup between the move-aside and the restore: that already exists — `UninstallUserDataAsync_WhenRestoreFailsAfterMoveAside_ReportsFailureAndKeepsTrackingDataAsync` deletes the backup inside the `CheckFileHashAsync` callback, which is exactly that window.

### One note for the team

DeepSource still flags the `catch (Exception ex) when (ex is A or B)` filtered form — it re-flagged the identical shape on HEAD after we narrowed it. Silencing it in the test helper would need four explicit catch clauses. I fixed the production occurrence and left the test helper alone rather than add that noise.

1664 passing; the only failures are the environment-dependent real-engine tests, which pass when run alone.

bobtista:
Both findings were real. The first one matters most: a guard added last round had reintroduced the exact corruption this PR exists to prevent.

### The same-path skip was preserving links to CAS objects

Last round `IsSamePath` was changed to resolve **both** sides through `ResolveLinkTarget(..., returnFinalTarget: true)`, to stop a copy overwriting its own source. Consequence: a destination symlink pointing at the CAS object compared equal to the CAS source, so `CopyFileAsync` returned before the unlink-and-copy. `HybridCopySymlinkStrategy` and every other `CopyFromCasAsync` caller ask for an *independent* copy — so the leftover link survived, and later writes through it reached the canonical CAS object. That is the corruption this PR was opened to stop, reintroduced by a fix for a different finding in the same file.

**The predicate turns on which side is the link**, because unlinking a link destroys nothing while unlinking a real file may destroy the only copy:

- **Identical normalized paths** → skip. Nothing to do, and unlinking would destroy the only copy. Covers the degenerate copy-a-symlink-onto-itself case too.
- **Destination is itself a link** → never skip, wherever it points. The caller wants a real file there; deleting a link leaves its target untouched, and the source is already open before the unlink, so even a link to the source is safe.
- **Destination is a real file** → skip only if the *source* resolves to it. Here the destination is the independent file the caller wants, and unlinking it would destroy content the source merely points at. This is the case the original guard existed for, and it is preserved.

`IsSamePath`/`ResolveFinalPath` became `WouldCopyOntoItself` plus `TryGetFullPath` (pure normalization) and `TryResolveLinkTarget`, which returns `null` when the path is not a link rather than falling back to the path itself — that fallback is precisely what made link-ness invisible to the caller. Hard links and Windows 8.3 short names still have no resolvable target and remain undetectable; the open-source-before-unlink ordering keeps those failing safely, and the doc comment says so.

Verified the regression test bites: restoring the old resolve-both-sides predicate makes `CopyFileAsync_DestinationIsSymlinkToSource_ReplacesLinkWithIndependentCopyAsync` fail while the other 23 pass. It asserts the link is gone, the content matches, and that writing to the destination afterwards leaves the source unchanged.

### The locked-backup test assumed the mode was enforced

Correct — `chmod 0500` is advisory under `CAP_DAC_OVERRIDE`, so under root the delete succeeds and the assertion fails as though the product were broken.

Rather than sniffing euid, the test now **proves** the denial: it writes a probe file into the backup directory before tightening the mode, then attempts to delete the probe afterwards. If that delete succeeds, the mode is advisory for this process and the test restores the mode and returns without asserting; otherwise the real assertions run and the probe is cleaned up in the `finally`. xunit 2.9.2 has no `Assert.Skip`, so this uses the silent early-return the codebase already uses for the symlink-privilege case.

Confirmed the probe is not vacuously skipping: with a temporary `Assert.Fail` in the privileged branch, the test still passes here — so the probe correctly observes that `0500` does deny the delete on this machine and the real assertions execute.

### Honest weakness

`CopyFileAsync_SourceIsSymlinkToDestination_LeavesFileIntactAsync` is weakly discriminating on Unix: with the skip removed entirely, macOS and Linux would unlink the destination while the stream is open and still reproduce the content, so it would pass anyway. It fails on Windows, where the sharing violation is real, and it documents intent. A cross-platform proof would need inode or hard-link inspection, which felt disproportionate for a guard test — flagging it rather than implying coverage it does not have.

1668 passing, 0 failing — the usually-flaky real-engine trio passed this run too.

undead2146 (approved):
LGTM

undead2146 (approved):
LGTM

PR #384, opened 2026-08-18
fix(config): preserve profiles, settings and workspace metadata across the v0.0.3 upgrade

bobtista:
## Problem

The v0.0.3 -> v0.0.4 upgrade path is broken in three independent ways. Together they mean an upgrading user loses their profiles and settings, and has every workspace rebuilt.

**1. The data root moved with no migration.** Four resolutions in `AppConfiguration` changed from `SpecialFolder.ApplicationData` to `SpecialFolder.LocalApplicationData` (Windows `%APPDATA%\GenHub` -> `%LOCALAPPDATA%\GenHub`; Linux `~/.config` -> `~/.local/share`). The only existing migration, `MigrateContentDirectory`, reshuffles a `Content` subdirectory *within* one root and returns early when it does not exist. So **every** alpha-3 user starts alpha 4 against an empty root and loses profiles, manifests, tracked user data, workspace metadata and settings.

**2. `WorkspaceStrategy` was renumbered.** v0.0.3 had `SymlinkOnly=0, FullCopy=1, HybridCopySymlink=2, HardLink=3`; current `main` has `HardLink=0, SymlinkOnly=1, ...`. Alpha-3 wrote `workspaces.json` with raw **ints** (`WorkspaceManager`'s options have no `JsonStringEnumConverter`) but wrote profiles as **strings** (`GameProfileRepository`'s options do). On upgrade the workspace side is misread by ordinal while the profile side parses correctly by name, so the two disagree, `WorkspaceManager`'s strategy-mismatch check sets `ForceRecreate`, and every existing multi-GB workspace is deleted and rebuilt.

**3. Profiles and manifests ignored the `ApplicationDataPath` override.** `GetProfilesPath()`/`GetManifestsPath()` used `GetConfiguredDataPath()` directly while `GetApplicationDataPath()`, `GetWorkspacePath()` and `GetCachePath()` all honor an explicitly-set override — so a user who relocated their data directory had profiles and manifests resolve somewhere else entirely.

## Changes

- **Restored the original enum ordinals** and added a remark that the numeric values are part of the on-disk format. `JsonWorkspaceStrategyConverter.Write()` now emits the member **name** so a future reordering cannot recur; `Read()` still accepts legacy numbers and strings.
- **New legacy data root migration** (`MigrateLegacyDataRoot`), run inside the existing migration lock ahead of `MigrateContentDirectory`. It no-ops when the legacy root is absent, when both roots resolve to the same path, or when the legacy root holds none of the tracked entries. It migrates exactly the settings file, `Profiles/`, `Manifests/`, `UserData/` and `workspaces.json`, each wrapped so a failure logs and continues rather than breaking startup, and never deletes the legacy root. `MigrateFile` skips any destination that already exists, which makes the whole thing idempotent and non-destructive. `MigrateDirectory` gained an `IOException` fallback to per-entry migration for the cross-volume case.
- **The CAS pool is deliberately excluded.** `GetCasConfiguration` still defaults `CasRootPath` to the legacy roaming location and alpha-3 used the same location, so CAS is currently consistent; moving the pool would orphan multi-GB stores.
- **`GetProfilesPath()`/`GetManifestsPath()`** now route through `GetApplicationDataPath()`.
- **Settings are correct on the first launch, not the second.** `UserSettingsService` builds in its constructor, before anything triggers the migration, so it would have read the empty new root, initialized defaults, and then overwritten the just-migrated file on the first save — destroying the very settings this PR exists to preserve, including `ApplicationDataPath` and `CasRootPath`. It now resolves its *read* path to the legacy file when the current root has none, while writes always target the new root; the move stays solely in the migration. `InitializeSettings` is also now failure-tolerant so it cannot throw out of the constructor.
- Replaced the `"Manifests"`/`"UserData"`/`"workspaces.json"` literals with constants.

## Tests

1638 -> 1681, all passing. Verified the new tests are genuine regression tests: reverting the settings resolution makes exactly the three sequence-specific tests fail.

- legacy `workspaces.json` with raw int strategies deserializes to the correct strategy; legacy profile strings still parse; converter round-trips and reads both encodings
- legacy workspace and profile metadata now **agree** — the comparison that triggered `ForceRecreate`
- `GetProfilesPath`/`GetManifestsPath` honor an explicit override
- migration moves the five tracked entries, is idempotent, does not overwrite existing files, leaves `CasPool` untouched, and no-ops without a legacy root or with identical roots
- first launch loads legacy settings; load -> migrate -> save preserves them; a legacy `ApplicationDataPath` override is honored immediately with no restart; second launch ignores the legacy file; fresh install is unaffected; a throwing legacy lookup still completes construction

## Notes

The migration **moves** rather than copies, matching the existing `MigrateDirectory` behavior, so a user who downgrades to v0.0.3 after upgrading will not find their data in the old location. Copying instead would double disk usage for what may be a large `UserData` tree. Worth a conscious call before release.

Corrected three pre-existing tests that hardcoded the post-renumbering ordinals, and added `AppConfiguration.cs` to the `ApplicationDataPathConventionTests` allowlist since it now legitimately reads the roaming folder to resolve the legacy root.

### DIFF
--- a/GenHub/GenHub.Core/Constants/DirectoryNames.cs
+++ b/GenHub/GenHub.Core/Constants/DirectoryNames.cs
@@ -51,6 +51,11 @@ public static class DirectoryNames
     public const string Profiles = "Profiles";
 
     /// <summary>
+    /// Directory holding manifests authored by the user, alongside <see cref="FileTypes.ManifestsDirectory"/>.
+    /// </summary>
+    public const string CustomManifests = "CustomManifests";
+
+    /// <summary>
     /// Directory that releases up to v0.0.3 nested the manifests, tracked user data and workspace
     /// metadata under. Current releases keep those entries directly in the data root.
     /// </summary>
--- a/GenHub/GenHub.Core/Constants/DirectoryNames.cs
+++ b/GenHub/GenHub.Core/Constants/DirectoryNames.cs
@@ -51,9 +51,33 @@ public static class DirectoryNames
     public const string Profiles = "Profiles";
 
     /// <summary>
+    /// Directory that releases up to v0.0.3 nested the manifests, tracked user data and workspace
+    /// metadata under. Current releases keep those entries directly in the data root.
+    /// </summary>
+    public const string LegacyContent = "Content";
+
+    /// <summary>
     /// Directory for storing tracked user data.
     /// </summary>
     public const string UserData = "UserData";
+
+    /// <summary>
+    /// Directory holding the manifests of tracked user data, nested inside <see cref="UserData"/>.
+    /// </summary>
+    /// <remarks>
+    /// Deliberately lower-case and separate from <see cref="FileTypes.ManifestsDirectory"/>: this is
+    /// the exact name written to disk, and matching case matters on case-sensitive filesystems.
+    /// </remarks>
+    public const string UserDataManifests = "manifests";
+
+    /// <summary>
+    /// Directory holding backups of replaced user data files, nested inside <see cref="UserData"/>.
+    /// </summary>
+    /// <remarks>
+    /// Deliberately lower-case and separate from <see cref="Backups"/>: this is the exact name
+    /// written to disk, and matching case matters on case-sensitive filesystems.
+    /// </remarks>
+    public const string UserDataBackups = "backups";
 
     /// <summary>
     /// Directory for storing workspaces.
--- a/GenHub/GenHub.Core/Constants/DirectoryNames.cs
+++ b/GenHub/GenHub.Core/Constants/DirectoryNames.cs
@@ -51,6 +51,11 @@ public static class DirectoryNames
     public const string Profiles = "Profiles";
 
     /// <summary>
+    /// Directory for storing tracked user data.
+    /// </summary>
+    public const string UserData = "UserData";
+
+    /// <summary>
     /// Directory for storing workspaces.
     /// </summary>
     public const string Workspaces = "Workspaces";
--- a/GenHub/GenHub.Core/Constants/FileTypes.cs
+++ b/GenHub/GenHub.Core/Constants/FileTypes.cs
@@ -36,6 +36,11 @@ public static class FileTypes
     public const string SettingsFileName = "settings.json";
 
     /// <summary>
+    /// File name holding the persisted workspace metadata.
+    /// </summary>
+    public const string WorkspaceMetadataFileName = "workspaces.json";
+
+    /// <summary>
     /// File extension for replay files.
     /// </summary>
     public const string ReplayFileExtension = ".rep";
--- a/GenHub/GenHub.Core/Constants/FileTypes.cs
+++ b/GenHub/GenHub.Core/Constants/FileTypes.cs
@@ -36,9 +36,20 @@ public static class FileTypes
     public const string SettingsFileName = "settings.json";
 
     /// <summary>
+    /// Settings file name written by releases up to v0.0.3, which combined the data root with the
+    /// JSON extension instead of the settings file name.
+    /// </summary>
+    public const string LegacySettingsFileName = ".json";
+
+    /// <summary>
     /// File name holding the persisted workspace metadata.
     /// </summary>
     public const string WorkspaceMetadataFileName = "workspaces.json";
+
+    /// <summary>
+    /// File name of the index tracking installed user data.
+    /// </summary>
+    public const string UserDataIndexFileName = "index.json";
 
     /// <summary>
     /// File extension for replay files.
--- a/GenHub/GenHub.Core/Helpers/PathHelper.cs
+++ b/GenHub/GenHub.Core/Helpers/PathHelper.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Security;
 
 namespace GenHub.Core.Helpers;
 
@@ -25,6 +26,28 @@ public static class PathHelper
         OperatingSystem.IsWindows()
             ? StringComparer.OrdinalIgnoreCase
             : StringComparer.Ordinal;
+
+    /// <summary>
+    /// Determines whether two paths point at the same filesystem location, normalizing both and
+    /// comparing them with the platform-appropriate case sensitivity.
+    /// </summary>
+    /// <param name="first">The first path.</param>
+    /// <param name="second">The second path.</param>
+    /// <returns><see langword="true"/> when both paths resolve to the same location.</returns>
+    public static bool AreSamePath(string first, string second)
+    {
+        try
+        {
+            return string.Equals(
+                Path.TrimEndingDirectorySeparator(Path.GetFullPath(first)),
+                Path.TrimEndingDirectorySeparator(Path.GetFullPath(second)),
+                PathComparison);
+        }
+        catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or SecurityException or NotSupportedException or ArgumentException)
+        {
+            return string.Equals(first, second, PathComparison);
+        }
+    }
 
     /// <summary>
     /// Gets the parent directory of a path, with fallback to the path itself if at drive root.
--- a/GenHub/GenHub.Core/Interfaces/Common/IAppConfiguration.cs
+++ b/GenHub/GenHub.Core/Interfaces/Common/IAppConfiguration.cs
@@ -12,6 +12,10 @@ public interface IAppConfiguration
     /// <returns>The root application data path.</returns>
     string GetConfiguredDataPath();
 
+    /// <summary>Gets the root application data path used by releases up to v0.0.3, which stored data under the roaming profile.</summary>
+    /// <returns>The legacy root application data path.</returns>
+    string GetLegacyConfiguredDataPath();
+
     /// <summary>Gets the default workspace path for GenHub.</summary>
     /// <returns>The default workspace path.</returns>
     string GetDefaultWorkspacePath();
--- a/GenHub/GenHub.Core/Models/Enums/WorkspaceStrategy.cs
+++ b/GenHub/GenHub.Core/Models/Enums/WorkspaceStrategy.cs
@@ -6,27 +6,31 @@ namespace GenHub.Core.Models.Enums;
 /// <summary>
 /// Workspace preparation strategy preference.
 /// </summary>
+/// <remarks>
+/// The numeric values are part of the on-disk format for workspace metadata written by
+/// releases up to v0.0.3 and must not be reordered.
+/// </remarks>
 [JsonConverter(typeof(JsonWorkspaceStrategyConverter))]
 public enum WorkspaceStrategy
 {
     /// <summary>
-    /// Hard link strategy - creates hard links where possible, copies otherwise. Space-efficient, requires same volume.
-    /// Default strategy for new profiles.
-    /// </summary>
-    HardLink = 0,
-
-    /// <summary>
     /// Symlink only strategy - creates symbolic links to all files. Minimal disk usage, requires admin rights.
     /// </summary>
-    SymlinkOnly = 1,
+    SymlinkOnly = 0,
 
     /// <summary>
     /// Full copy strategy - copies all files to workspace. Maximum compatibility and isolation, highest disk usage.
     /// </summary>
-    FullCopy = 2,
+    FullCopy = 1,
 
     /// <summary>
     /// Hybrid copy/symlink strategy - copies essential files, symlinks others. Balanced disk usage and compatibility.
     /// </summary>
-    HybridCopySymlink = 3,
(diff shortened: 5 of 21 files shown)


bobtista:
The DeepSource review flags two things here. Both are pre-existing on `development` and outside this PR's diff.

**`settingsToSave` reported as uninitialized** — `UserSettingsService.cs:149`. It is declared and then assigned unconditionally inside the `lock` block immediately below, so C# definite-assignment holds and the file compiles. That makes it a false positive, and the code is untouched by this branch — byte-identical on `development`.

**Generic `catch` in `ConfigurationProviderService`** — the three unfiltered `catch (Exception ex)` blocks left on this branch (lines 60, 90, 488) are pre-existing; they sit at lines 45, 75 and 387 on `development`. Every catch this PR *adds* is already constrained by an exception filter:

```csharp
catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or SecurityException or NotSupportedException or ArgumentException)
```

That was the earlier inline finding on this file, addressed in 14df2c30. The broad-but-filtered form is deliberate for the migration paths: a failure there must log and continue rather than break startup.

Note the analyzer also reports `DeepSource: C#` as failing on `development` itself at `495bcc8c`, so this check is not a signal introduced by this PR. Happy to clean up the pre-existing catches separately — I'd rather not widen an upgrade-path fix to do it.

bobtista:
Second round addressed. The CRITICAL was real, and chasing it surfaced two further defects that would have made this PR fail at its core job for actual v0.0.3 upgraders.

### The reported CRITICAL — real, fixed

`MigrateLegacyDataRoot` used `GetConfiguredDataPath()` as the destination while `GetProfilesPath()`/`GetManifestsPath()` read from `GetApplicationDataPath()`. For a user with an `ApplicationDataPath` override, the migration **moved** their data into the non-override root while the app read the override root — worse than not migrating, since the legacy copy was gone too.

Now takes three roots: data (override-aware) for `Profiles`/`Manifests`/`UserData`/`workspaces.json`, and the configured root for `settings.json`, because `UserSettingsService` writes there and sending it to the override root would strand it — the same failure mirrored. The ordering has no cycle: `UserSettingsService` resolves its path from `IAppConfiguration` alone and never calls the configuration provider, and its existing legacy fallback means the override from the *legacy* settings file is already in memory before migration runs. `GetApplicationDataPath()` was split so the migration can read the override without re-entering itself.

### Two defects found while verifying, both verified against the `v0.0.3` tag

**v0.0.3 wrote the settings file as `.json`, not `settings.json`.** At that tag `GetDefaultSettingsFilePath()` was `Path.Combine(GetConfiguredDataPath(), FileTypes.JsonFileExtension)` — the extension constant used as a filename. It was corrected to `FileTypes.SettingsFileName` in #230, which post-dates v0.0.3. So the legacy lookup this PR added would never have found a real v0.0.3 user's settings, defeating the headline feature. Both names are now probed, current name winning, and always landing under the current name.

**v0.0.3 nested manifests, user data and workspace metadata under `Content/`.** `GetApplicationDataPath()` returned `<root>/Content` at that tag, and `MigrateContentDirectory` did not exist yet — it arrived in `a7f2e4ce` (#230), after the tag. So the flat-only probe would have migrated `Profiles/` and settings while silently abandoning manifests, tracked user data and workspace metadata for the primary target population. `MigrateContentDirectory` could not rescue them either, since it probes `<newRoot>/Content` under local appdata while theirs sits under roaming. Both layouts are now probed, flat first so a newer copy wins.

### Other fixes

- **Failed initialization can no longer overwrite real settings.** A `_settingsLoaded` flag makes `SaveAsync` refuse rather than write defaults to a path that may hold real settings. Throwing suits the existing contract (`SaveAsync` already throws IO/JSON exceptions) and `TryUpdateAndSaveAsync` converts it to `false`.
- **Normalization no longer destroys a good load** — a `Math.Clamp` throw from bad appsettings kept the loaded values instead of replacing them with defaults.
- **Cross-volume migration actually works** — the `IOException` fallback recursed into a `File.Move`-only helper that would fail on the same condition; it now copies then deletes, and per-entry failures no longer abort the whole directory.
- **Path comparisons centralized** — extracted `PathHelper.AreSamePath` and removed the hardcoded `OrdinalIgnoreCase` comparisons from `UserSettingsService`; remaining broad catches narrowed to the set established in `14df2c30`.
- Exact-cased constants for the user-data sub-paths, corrected provenance docs, `docs/dev/constants.md` updated, and coverage for the no-entries early return plus negative assertions so a move→copy regression cannot pass idempotency.

### Rejected

The claim that v0.0.3 flattened `Content/` on every launch is wrong — `MigrateContentDirectory` post-dates the tag, as above. Also skipped driving a workspace-metadata test through the real `WorkspaceManager` read path: that path is option-less today and the converter is applied via the enum attribute, so the divergence it guards against cannot occur.

### Known gap, needs a release note rather than code

At v0.0.3, `GameProfileModule` computed the profiles directory as `Path.GetDirectoryName(GetApplicationDataPath())/Profiles`. With an override of, say, `D:\GenHubData` that landed on `D:\Profiles` — a v0.0.3 bug. Those profiles sit outside the legacy root entirely, so this migration cannot reach them without probing arbitrary parent directories. Relocated-data alpha-3 users may need to move that folder by hand.

1688 passing (from 1679); the only failures are the environment-dependent real-engine tests.

bobtista:
Third round. All three findings were real; one of them partially defeated a fix from the previous round, so it is worth being explicit about.

### `GetContentDirectories` bypassed the override — same defect class as the earlier CRITICAL

`GetManifestsPath()` resolves through `GetApplicationDataPath()` and the legacy migration moves `Manifests` into the override root, but `GetContentDirectories()` still built its defaults from `GetConfiguredDataPath()`. For a user with `ApplicationDataPath` set, local content discovery scanned a directory holding no manifests while reads and writes went elsewhere. Both defaults now resolve from one `GetApplicationDataPath()` call, and `"CustomManifests"` moved into `DirectoryNames` (the one other bare occurrence, in `ManifestDiscoveryService`, was already correctly rooted and now uses the constant).

Since this is the second time this exact split has appeared, here is the full audit of every `GetConfiguredDataPath()` use rather than another one-off fix:

- `GetContentDirectories` — the bug, fixed.
- `GetRootAppDataPath()` — **left deliberately.** It is the installation root by definition; its only production consumer opens the directory holding the settings file. Making it override-aware would make it a duplicate of `GetApplicationDataPath()`.
- `ResolveApplicationDataPath()` — left, it *is* the no-override fallback.
- `MigrateLegacyDataRoot()` — left. Data root is override-aware, settings root is not, which is last round's deliberate decision because `UserSettingsService` writes there.
- `GetCasConfiguration()` / `GetLogsPath()` — left, both use `SpecialFolder` directly and the CAS exclusion is documented in the migration remark.
- Outside this file, `MapPackService`, `UploadHistoryService` and `ContentPipelineModule` root at the configured path but none is read back through `GetApplicationDataPath()`, so none has the split-root defect. Whether map packs *should* follow the override is a product question, left alone.

### A swallowed load failure was still marked as loaded

This is the one that partially undid last round's work. `LoadSettings` swallows `IOException`, `UnauthorizedAccessException` and `JsonException` internally and returns defaults; `InitializeSettings` then set `_settingsLoaded = true` regardless. So a corrupt or locked settings file produced defaults that the next `SaveAsync` would write over the user's real file — exactly the data loss the `_settingsLoaded` guard was added to prevent. The guard only covered exceptions that escaped to the outer catch, not the ones `LoadSettings` ate.

`LoadSettings` now reports `Absent` / `Loaded` / `Failed`, and `_settingsLoaded` is set only for the first two, using the outcome of whichever load actually produced the settings — so a corrupt *legacy* file blocks saving too, via the same call path.

Two judgment calls worth stating: a missing file and an empty/whitespace file are both `Absent`, because neither holds anything a save could destroy and treating a zero-byte file as `Failed` would wedge that user permanently; a file that deserializes to `null` or throws is `Failed`.

### `_migrated` is now `volatile`

Plain `bool` read on the fast path outside the lock. On ARM64 — every Apple Silicon Mac — a thread seeing `true` was not guaranteed to see the completed migration, so it could read profiles or manifests mid-migration. The write is the last statement inside the lock, after both migrations complete, so the release semantics publish all of it.

### Tests

1688 -> 1693. Four new regression tests cover the direction that matters: a corrupt file in the current root, a corrupt legacy file, and a file held with `FileShare.None` all refuse the save and leave the original bytes intact; and a first run with no file at all still persists normally, so the block does not over-trigger.

undead2146 (dismissed):
LGTM

bobtista:
This one was user-reachable data destruction, and it is the fourth time in this series that a guard I added failed to cover a path reaching the same operation. So it is fixed structurally rather than at the two known call sites.

### The bug

`Update()` and `TryUpdateAndSaveAsync()` re-pointed `_settingsFilePath` to `_settings.SettingsFilePath` without loading that target and without touching `_settingsLoaded`. The guard then read "loaded" — from a different file — and the save proceeded. Reachable from the UI: `SettingsViewModel.SaveSettings` takes the path from an editable field or the browse picker, so pointing GenHub at an existing settings file overwrote it with in-memory state derived from something else.

### Every path that could reach it

Mutating the write path: the `initialize: false` constructor, `InitializeSettings`, `SetSettingsFilePath` (the only correct one — it loads), and the two bypasses above.

Reaching the write: `SaveAsync` from `SettingsViewModel` (save and reset-to-defaults), `App.axaml.cs` shutdown, `MainViewModel` ×2, `ToolService` ×2, `UpdateNotificationViewModel` ×5; and `TryUpdateAndSaveAsync` from `NotificationService` ×3, `StorageLocationService`, `InstallationCasPoolService`, and three profile reconcilers ×2 each.

Only `SettingsViewModel` assigns `SettingsFilePath`, so it is the sole user-reachable trigger — but roughly eighteen save sites share one write target, which is why this had to be fixed in the state rather than at the callers.

### The structural change

`_settingsFilePath` and `_settingsLoaded` are replaced by a single immutable `SettingsFileTarget` holding `Path` (where a save writes) and `VerifiedPath` (the file last verified safe to overwrite), with `CanWrite` requiring the two to agree.

Its constructor is private; the only ways to get one are `Verified`, `Unverified` and `MoveTo` — and **`MoveTo` carries the verified file, not the permission**, so moving the write path automatically drops writability unless it lands back on the verified file. There is no way to move the path and leave a stale "already read" flag, because they are one value and the compiler makes a caller state which factory it means. A future call site that re-points has three options, none of which silently permits a destructive write.

The `SettingsLoadOutcome → target` mapping now lives in one place, so the "Absent counts as safe" rule for first run and legacy upgrade cannot drift between the three load sites. `SetSettingsFilePath` also now takes the lock, which it did not despite mutating shared state.

### What happens when a user re-points

- **Target holds nothing** (missing or empty) → adopted, edits kept, save proceeds. The supported relocate flow, and what the Save-file picker produces.
- **Target already holds settings**, readable or not → the path moves but stays unverified, so the save refuses. Nothing is destroyed and nothing is discarded: the file keeps its contents and the user keeps every edit.

I rejected *reload* because it silently throws away the edits the user just clicked Save on, and *overwrite* because it is the bug. Refusal is the only reading that loses nothing, and it leaves the genuine ambiguity — "start using this file" versus "save my settings there" — for the user to resolve.

The refusal is not sticky and cannot strand the app, and that fell out of the design rather than needing another flag: pointing back at the original file re-verifies, and re-pointing once the conflicting file is gone re-probes and adopts. Both covered by tests.

**On routing both sites through `SetSettingsFilePath`:** deliberately not done. That method is the "start using this file" operation and by design replaces the in-memory settings with the file's contents — routing in-flight edits through it would implement exactly the silent discard I rejected. They are different operations; what they need to share is the invariant, which they now do.

### Also

`SettingsViewModel.SaveSettings` swallowed every save failure into a log line, so a refusal would have looked like nothing happened. It now raises an error notification — refusing without telling the user is its own problem.

Replaced two private-field reflection hacks in existing tests with a real seam. They would have broken on the removed field anyway, and reflecting into private state is what let this class of bug hide.

### Tests

Seven new, covering both bypasses, the legitimate relocate, both un-stranding paths, settings that name their own file, and first run. Verified the first four **fail** against the old adopt-unconditionally behaviour.

1700 passing; the 3 failures are the environment-dependent real-engine tests.

### Flagged, not fixed

- `ResetToDefaults` sets `SettingsFilePath` empty, and a blank path leaves the target alone — so it does not move settings back to the default location. Pre-existing and unchanged, but a latent surprise.
- `UpdateNotificationViewModel` calls `_ = SaveAsync()` in five places: fire-and-forget with no continuation, so any failure including this refusal is an unobserved faulted task. That is a real hole in the "refuse loudly" story, out of scope here.

undead2146 (approved):
LGTM

PR #385, opened 2026-08-18
fix(content): bound archive extraction and propagate cancellation from deliverers

bobtista:
## Problem

Five defects in archive handling, across both remote content delivery and local imports.

**1. A cancelled GitHub extraction reported success.** `ExtractArchiveAsync` handled cancellation with a mid-loop `break`. The `Task.Run(..., cancellationToken)` wrapper only honors the token *before* the delegate starts, so the break completed the task successfully — the caller then logged success, **deleted the downloaded archive**, and registered the truncated file set as a complete manifest. Unrecoverable locally, because the archive was already gone.

**2. `PathSeparators` contained an apostrophe, not a backslash.** `['/', '\'']` — an escaped `'` (0x27) where `'\\'` was intended. Backslash-separated zip entries were therefore never split, so the segment-based traversal check could not see `..\` segments, and a legitimate name containing an apostrophe (`Bob's Map/bob.map`) was mis-split and mis-grouped.

**3. Expansion limits trusted attacker-declared sizes.** The map and replay validators checked `entry.Length` / `entry.CompressedLength` — the ZIP central-directory *declared* values — then extracted with unbounded `ExtractToFile`/`CopyTo`.

**4. Community Outpost extraction had no containment check or expansion bound of its own.** It called `WriteToDirectory` with `ExtractFullPath = true` on archives fetched from a third-party catalog and its mirrors, relying on undocumented library behavior for traversal safety, with no size or count caps.

**5. SharpCompress 0.42.0** is subject to [GHSA-6c8g-7p36-r338](https://github.com/advisories/GHSA-6c8g-7p36-r338) (`NU1902` is raised on every build today).

### Verified while fixing

`System.IO.Compression`'s `ZipArchiveEntry.Open()` **clamps** reads at the declared uncompressed size — a zip patched to declare 4 KB while holding 12 MB yields exactly 4096 bytes. **SharpCompress does not clamp**: the same archive inflates the full 12,582,912 bytes. So the header-trust hole is genuinely exploitable on the SharpCompress deliverers (GitHub, Community Outpost), while on the `System.IO.Compression` importers the bounded copy is defense in depth plus a real fix for asset amplification, where an asset shared by N maps is written N times but counted once.

## Changes

- **Cancellation** — `ThrowIfCancellationRequested()` replaces the `break`, with `catch (OperationCanceledException) { throw; }` ahead of the per-archive and method-level handlers so the archive is kept and nothing reaches the manifest pool.
- **Shared helpers, three duplicates removed.** `BoundedArchiveExtractor.CopyEntryToFileAsync` (in `GenHub.Core/Utilities`, alongside the existing `ZipValidation`) streams an entry against `min(perEntryCap, remainingAggregate)`, aborts mid-stream, deletes the partial output and throws `ArchiveExpansionLimitExceededException`. It takes a plain `Stream`, so Core needs no SharpCompress reference and both `System.IO.Compression` and SharpCompress callers share it. The containment check moved into the existing `PathHelper.IsPathWithinDirectory`, replacing copies in `GitHubContentDeliverer` and `ContentStorageService`.
- **All four extraction paths are now bounded and containment-checked**: map import, replay import, Community Outpost (both the main package and the dependency-archive call site), and GitHub.
- **Typo fixed** to `['/', '\\']`, with the surrounding grouping and validation re-checked for both separators now that splitting actually works.
- **SharpCompress 0.48.0.** Not drop-in: `ArchiveFactory.Open` was renamed to `ArchiveFactory.OpenArchive` (the async surface took the old names). Two call sites updated; `NU1902` is gone.

## Caps

GitHub caps a single release asset at 2 GiB, which anchors the numbers. GitHub uses `min(16 GiB, max(8 MiB, archiveSize × 500))` — a ratio-derived budget matters because absolute caps alone do nothing against a 12 KB bomb. Real deflate release content runs 2–20:1 against a 500:1 limit.

Community Outpost deliberately keeps **absolute** caps only (2 GiB entry / 4 GiB aggregate / 10,000 entries) rather than a ratio. Its payloads are 7z/LZMA over BIG files with internal padding, where the realistic ceiling is not predictable and a badly chosen ratio would reject legitimate catalog content in the field. Happy to mirror the ratio there if you'd rather.

## Tests

1638 -> 1660, fully green. Confirmed discriminating: reverting each source fix fails exactly the intended tests.

- containment helper accepts contained paths and rejects `..`, traversal, normalized-escape, sibling-prefix and absolute-outside paths
- bounded copy honors per-entry and aggregate budgets, deletes partial output, and rejects an archive whose central directory and local header are patched to declare 4 KB while inflating to 12 MB
- backslash traversal rejected; backslash-separated entries resolve to one directory; apostrophes in names stay intact
- Community Outpost rejects an escaping entry and an over-budget archive
- GitHub: cancelling mid-extraction propagates, keeps the archive and registers nothing; a size-understating archive fails **without** registering a manifest and is distinguishable from cancellation

### DIFF
--- a/GenHub/Directory.Packages.props
+++ b/GenHub/Directory.Packages.props
@@ -47,7 +47,7 @@
     <PackageVersion Include="coverlet.collector" Version="6.0.2" />
     <PackageVersion Include="moq" Version="4.20.72" />
     <PackageVersion Include="FluentAssertions" Version="6.12.1" />
-    <PackageVersion Include="SharpCompress" Version="0.42.0" />
+    <PackageVersion Include="SharpCompress" Version="0.48.0" />
     <PackageVersion Include="System.Management" Version="10.0.2" />
     <!-- Image processing for AVIF to TGA conversion (GenPatcher content) -->
     <PackageVersion Include="SixLabors.ImageSharp" Version="3.1.12" />
--- a/GenHub/GenHub.Core/Constants/CommunityOutpostConstants.cs
+++ b/GenHub/GenHub.Core/Constants/CommunityOutpostConstants.cs
@@ -85,6 +85,22 @@ public static class CommunityOutpostConstants
     /// </summary>
     public const string PatchPageUrl = "https://legi.cc/downloads/genpatcher/";
 
+    /// <summary>
+    /// Maximum number of file entries a downloaded Community Outpost archive may contain.
+    /// </summary>
+    public const int MaxArchiveEntries = 10000;
+
+    /// <summary>
+    /// Maximum number of bytes a single Community Outpost archive entry may expand to (2 GB),
+    /// sized to accommodate the largest shipped BIG files.
+    /// </summary>
+    public const long MaxEntryUncompressedBytes = 2L * 1024 * 1024 * 1024;
+
+    /// <summary>
+    /// Maximum aggregate uncompressed bytes a Community Outpost archive may expand to (4 GB).
+    /// </summary>
+    public const long MaxAggregateUncompressedBytes = 4L * 1024 * 1024 * 1024;
+
     /// <summary>Display name for Game Clients content type.</summary>
     public const string ContentTypeGameClients = "Game Clients";
 
--- a/GenHub/GenHub.Core/Constants/CommunityOutpostConstants.cs
+++ b/GenHub/GenHub.Core/Constants/CommunityOutpostConstants.cs
@@ -91,13 +91,13 @@ public static class CommunityOutpostConstants
     public const int MaxArchiveEntries = 10000;
 
     /// <summary>
-    /// Maximum number of bytes a single Community Outpost archive entry may expand to (2 GB),
+    /// Maximum number of bytes a single Community Outpost archive entry may expand to (2 GiB),
     /// sized to accommodate the largest shipped BIG files.
     /// </summary>
     public const long MaxEntryUncompressedBytes = 2L * 1024 * 1024 * 1024;
 
     /// <summary>
-    /// Maximum aggregate uncompressed bytes a Community Outpost archive may expand to (4 GB).
+    /// Maximum aggregate uncompressed bytes a Community Outpost archive may expand to (4 GiB).
     /// </summary>
     public const long MaxAggregateUncompressedBytes = 4L * 1024 * 1024 * 1024;
 
--- a/GenHub/GenHub.Core/Constants/GitHubConstants.cs
+++ b/GenHub/GenHub.Core/Constants/GitHubConstants.cs
@@ -333,6 +333,33 @@ public static class GitHubConstants
     /// <summary>Description for GitHub content deliverer.</summary>
     public const string GitHubDelivererDescription = "Delivers GitHub content including release archives";
 
+    // Archive extraction limits
+    // GitHub caps a single release asset at 2 GiB, so a downloaded archive can never exceed that
+    // compressed. These bounds leave generous headroom above real game content while keeping an
+    // archive that lies about its declared sizes from expanding without limit.
+
+    /// <summary>Maximum number of file entries a downloaded GitHub archive may contain.</summary>
+    public const int MaxArchiveEntries = 50000;
+
+    /// <summary>Maximum number of bytes a single GitHub archive entry may expand to (4 GiB).</summary>
+    public const long MaxEntryUncompressedBytes = 4L * 1024 * 1024 * 1024;
+
+    /// <summary>Maximum aggregate uncompressed bytes a GitHub archive may expand to (16 GiB).</summary>
+    public const long MaxAggregateUncompressedBytes = 16L * 1024 * 1024 * 1024;
+
+    /// <summary>
+    /// Maximum factor by which a GitHub archive may expand beyond its own downloaded size. Release
+    /// archives are deflate-compressed game content and executables, which run well under 20:1, so
+    /// this bounds a small archive that claims to hold very little and then inflates without end.
+    /// </summary>
+    public const long MaxArchiveExpansionRatio = 500;
+
+    /// <summary>
+    /// Floor for the ratio-derived expansion budget (8 MiB), so a very small archive still gets
+    /// room for content that compresses unusually well and is judged only by the absolute caps.
+    /// </summary>
+    public const long MinArchiveExpansionBudgetBytes = 8L * 1024 * 1024;
+
     // Metadata keys
 
     /// <summary>Metadata key for repository owner.</summary>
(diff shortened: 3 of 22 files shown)


bobtista:
Triaged the bot findings. All were real and are now fixed, though one is not a regression from this PR — details below.

### Cancellation reported as success (`27f5d997`)

The three sibling paths were converting cancellation into ordinary success or failure, which contradicted this PR's own goal after `GitHubContentDeliverer` was fixed:

- `ReplayImportService.ImportFromZipAsync` — the outer `catch (Exception)` swallowed the cancellation and returned a normal `ImportResult`, so a truncated import reported `Success = true` whenever an earlier entry had landed.
- `MapImportService.ImportFromZipAsync` — same shape, with `result.Success = result.FilesImported > 0` reporting a cancelled partial import as successful.
- `CommunityOutpostDeliverer` — both the extraction-level and method-level catches turned it into `CreateFailure("Extraction failed: The operation was canceled")`, so callers could not tell user cancellation from a hostile or broken archive.

Worth noting: rethrowing at the archive level alone was not sufficient. The per-file catches in `ImportFromFilesAsync` / `ImportFromStreamAsync` and the per-dependency catch in the Community Outpost auto-install loop re-swallowed it one level up, so those needed to exclude cancellation too.

### Bounded extractor correctness (`26c10c42`)

- `read` initialized at declaration.
- An exhausted budget is now rejected before the file is opened. Previously `limit = Math.Min(maxEntryBytes, remainingAggregateBytes)` could be `<= 0` while the check only ran *inside* the read loop, so a zero-byte entry never entered the loop and returned success with the aggregate already spent.
- Overwrite no longer truncates in place. The entry is staged beside its destination with `FileMode.CreateNew` and moved into place only once the copy completes, so a failed replacement leaves the pre-existing file intact and the failure path deletes only what this call wrote. Non-overwrite behaviour is unchanged.

### Map import resilience (`570f842f`)

Both fixed, with one correction to the finding:

- **"Non-expansion errors abort the entire import" is accurate but is not a regression here.** Before this PR the code was a bare `mapEntry.ExtractToFile(mapDestPath, false)` with no per-map `try`/`catch` at all, so an `IOException` already aborted the whole archive. This PR introduced the first per-map handler; it narrowed nothing. Visible in `git show 882d57f0 -- GenHub/GenHub/Features/Tools/MapManager/Services/MapImportService.cs`, hunk `@@ -290,38 +293,67`. Widened it anyway to skip only the offending map, since it is a one-line change in a block already being touched.
- The aggregate counter is no longer charged for a discarded map: per-map bytes are folded into the running total only when the map survives, while in-flight budget still subtracts them so nothing is double-counted.

### Tests

1660 -> 1667, all passing. Each cancellation test fails against the pre-fix code specifically because the old path returned a normal result instead of throwing: cancelled map and replay imports no longer report success, cancelled Community Outpost delivery keeps the archive and registers nothing, a failed overwrite leaves the original file readable with no staging file behind, and an empty entry is rejected once the aggregate budget is spent.

bobtista:
Second round addressed, 7 commits. Two of the findings turned out to be broader than reported, and one suggestion I'm declining with reasoning.

### Entry names — broader than the report

The finding named empty/null `entry.Key`. `ZipArchive.CreateEntry("")` throws, so a truly empty key cannot be produced by .NET's writer — but `.`, `patch/..` and `" "` are all constructible and collapse identically: `Path.GetFullPath(Path.Combine(extractPath, "."))` **is** `extractPath`, and containment passes because the relative path is `.`. My first version of the guard allowed exactly those, and the test caught it.

New `ArchiveEntryName.IsExtractable` runs before any path is built in both deliverers. It refuses null/empty/whitespace, trailing separators, a final segment of `.` or `..`, trailing dot or space, Windows-invalid characters (including `:`, which would open an NTFS alternate data stream) and reserved device names. Interior `..` is still the containment check's job. The `entry.Key ?? string.Empty` fallbacks are gone.

### The cancellation test did not test cancellation — confirmed

I verified it the way you'd want: removed the per-entry `ThrowIfCancellationRequested` and passed `CancellationToken.None` to the copy, leaving only the outer token on `Task.Run`. **The old test still passed.** The `Extracting` progress report fires before `ExtractArchiveAsync`, so the task was cancelled before the loop body ever ran, and the `Length < entryCount` assertion held vacuously against an empty directory.

Rewritten so cancellation lands mid-entry: two tiny entries plus a 32 MiB one, with a watcher that cancels the moment `marker.dat` appears, so the token trips while `large.dat` is being written. It asserts `first.dat` exists (extraction genuinely started) and `large.dat` does not. Re-running the same revert against the new test **fails**, which is the property the old one lacked.

### Containment, and a note for #386

`IsPathWithinDirectory` keeps the exact lexical check and now *additionally* requires containment after link resolution.

Worth flagging for anyone touching the equivalent check elsewhere: a leaf-only `ResolveLinkTarget` is **not** safe here, because the candidate normally does not exist yet while the base does. Resolving one side but not the other would reject every extraction on macOS, where `/var` resolves to `/private/var`. The helper therefore walks to the deepest existing ancestor, resolves that, and re-appends the remainder, so both sides go through identical resolution. It never throws — IO/access/not-supported/argument failures return the path as written.

### Declining: the map-discard budget release

Not reachable by any archive a test can build, and I'd rather say so than write a test that only appears to cover it. `MapImportService.ValidateZip` already refuses an archive whose *declared* aggregate exceeds `MaxAggregateUncompressedBytes`, so declared ≤ 200 MiB. Triggering the runtime rejection needs `bulk + doomed > 200 MiB` while passing validation needs `bulk + doomed + survivor ≤ 200 MiB` — a direct contradiction for honest entries. Only a size-spoofing archive reaches it, and because the discarded map's release is capped at `MaxMapSizeBytes` (10 MiB), the surviving maps would still have to materialise ≥190 MiB on disk, on top of a new multi-entry spoofed-size fixture. That is ~200 MiB of I/O per suite run to cover a path that honest input cannot reach. I covered the cheap deterministic half instead: the aggregate budget shrinking across entries.

### Everything else taken

- Spent-budget failures get their own message instead of claiming the entry "expanded past the allowed 0 bytes"; both messages keep "potential zip bomb" so existing assertions hold.
- Staging name is now `GetRandomFileName() + suffix` in the destination's directory rather than destination + suffix + 32-char GUID, which was adding ~48 characters toward MAX_PATH.
- Cleanup moved into a quiet dispose so a failure there can no longer mask the original exception.
- `Directory.CreateDirectory` moved inside the per-map handler — verified the new test fails when reverted.
- GiB units corrected, `IoConstants` documented, and the fixture builder now asserts the EOCD, central-directory and local-header signatures before patching, so a layout drift fails loudly instead of silently corrupting the archive.

Eight older comments no longer hold: `read` uninitialized (`int read = 0;`), failed-overwrite-deletes-original (staging), Replay and Map cancellation reporting success (both rethrow OCE ahead of the general catch), Community Outpost cancellation propagation, zero-byte entries bypassing a spent budget (`limit <= 0` throws before the loop), non-expansion errors aborting the import, and the aggregate counter rollback.

1711 passing (+47); the only failures are the environment-dependent real-engine tests, which touch no archive or path-helper code.

undead2146 (dismissed):
LGTM

undead2146 (dismissed):
LGTM

PR #386, opened 2026-08-18
fix(launching): adopt a forked game process on Unix as well as Windows

bobtista:
## Problem

When a launched process exits 0 almost immediately, `HandleImmediateProcessExitAsync` looks for the real game process the launcher spawned and adopts it. That adoption was gated to Windows:

```csharp
if (exitCode == 0 && OperatingSystem.IsWindows())
```

So on macOS and Linux, a client that forks and exits is reported as `"Process exited immediately after launch."` while the game is actually running. `InstallationSourceConstants.Wine` is a recognized installation source, so Wine/Proton wrappers hit exactly this path.

## Why the gate is safe to remove

The gate arrived with the native BGFX launch work (`c2202bef`), where the concern was that a native engine exiting 0 would be falsely reported as running. That concern is already covered elsewhere:

- `GameProcessSelector.SelectSpawnedGameProcess` requires a name match, a start time at or after the launcher's, a recency window, **and** that the candidate's executable resides in the exact workspace working directory — "a same-named process elsewhere on the machine is somebody else's". A stray process cannot be adopted.
- If the engine really did exit, there is no second same-named process running from that workspace, the selector returns null, and the launch still fails loudly with the captured stderr. The gate was redundant for the case it was written for.
- The declared-child case is a different path: when `ExpectedChildProcessName` is set, `GameProcessManager` routes to `AdoptExpectedChildProcessAsync`, which was already cross-platform and is untouched here.

The residence check genuinely works on Unix — `Process.MainModule?.FileName` resolves for other processes on macOS (verified across 650 running processes: 643 resolved, zero unhandled exceptions) and Linux reads `/proc`.

Other `OperatingSystem.IsWindows()` checks in the file are deliberately left alone: the `.exe` extension fallback, and the execute-bit checks that are correctly Windows-exempt.

## Tests

`GameProcessSelectorTests` — the `launcherStartTime` filter was not covered and is now load-bearing on Unix, so it gains `RejectsCandidatesStartedBeforeTheLauncher` and `AcceptsACandidateStartedAtTheLauncherStartTime` (the `>=` boundary, since a child can be timestamped within the launcher's tick).

`GameProcessManagerTests` — `StartProcessAsync_WhenAnUndeclaredLauncherForksAndExits_AdoptsTheSpawnedGameAsync` drives the real path: a launcher that backgrounds a same-named child in the workspace and exits 0. It skips on Windows, kills the adopted process in a `finally`, and the harness sweeps stragglers. Confirmed load-bearing — restoring the Windows gate makes it fail with "Process exited immediately after launch."

1638 passing, stable across two consecutive full runs. The 3 failures are the pre-existing environment-dependent real-engine tests, unchanged by this fix.

## Two incidental findings

**`StartProcessAsync_WithExpectedChild_TracksTheChildWhileTheLauncherStillRunsAsync` was skipped on non-Windows for a false reason.** Its comment claimed `Process.GetProcessesByName` does not enumerate these processes on macOS. It does. The real cause was the fixture: a copy of `/bin/sleep` is SIGKILLed on Apple Silicon (exit 137) because copies of platform binaries fail code-signature validation. The harness now ad-hoc signs the copy on macOS and probes that the child runs, skipping rather than failing if a platform still refuses. The test is un-gated and passes.

**`GameProcessSelector` residence comparison does not resolve symlinks.** `MainModule.FileName` is fully resolved (`/private/var/folders/...`) while an unresolved path may contain a symlinked component (`/var/folders/...`), and `Path.GetFullPath` does not collapse that — so residence never matches and adoption silently fails. I canonicalize the path in the test harness rather than change the selector, since that is out of scope here. **Worth a follow-up:** on macOS/Linux a workspace reached through a symlinked component will silently defeat this check. Real workspaces under `~/Library/Application Support` are unaffected, but a user with a symlinked data directory would be.

### DIFF
--- a/GenHub/GenHub.Core/Constants/IoConstants.cs
+++ b/GenHub/GenHub.Core/Constants/IoConstants.cs
@@ -9,4 +9,10 @@ public static class IoConstants
     /// Default buffer size for file operations (4KB).
     /// </summary>
     public const int DefaultFileBufferSize = 4096;
+
+    /// <summary>
+    /// How many times a path may be re-resolved while following symbolic links whose targets are
+    /// themselves reached through links. Bounds the walk on a filesystem that contains a cycle.
+    /// </summary>
+    public const int MaxSymbolicLinkResolutionDepth = 8;
 }
--- a/GenHub/GenHub.Core/Constants/ProcessConstants.cs
+++ b/GenHub/GenHub.Core/Constants/ProcessConstants.cs
@@ -84,6 +84,13 @@ public static class ProcessConstants
     public const double EarlyExitThresholdSeconds = 10.0;
 
     /// <summary>
+    /// How many characters of a process name a Unix kernel keeps. Linux stores it in a
+    /// TASK_COMM_LEN buffer and macOS in a MAXCOMLEN one, both of which leave room for fifteen
+    /// characters and a terminator, and the truncated value is what process enumeration matches on.
+    /// </summary>
+    public const int UnixProcessNameMaxLength = 15;
+
+    /// <summary>
     /// How long to wait for a launcher's expected child process to appear. Measured spawn latency
     /// for the Easy Anti-Cheat bootstrapper is well under two seconds. Must not exceed
     /// <see cref="EarlyExitThresholdSeconds"/>, which bounds how old an adoptable process may be.
--- a/GenHub/GenHub.Core/Constants/ProcessConstants.cs
+++ b/GenHub/GenHub.Core/Constants/ProcessConstants.cs
@@ -92,8 +92,9 @@ public static class ProcessConstants
 
     /// <summary>
     /// How long to wait for a launcher's expected child process to appear. Measured spawn latency
-    /// for the Easy Anti-Cheat bootstrapper is well under two seconds. Must not exceed
-    /// <see cref="EarlyExitThresholdSeconds"/>, which bounds how old an adoptable process may be.
+    /// for the Easy Anti-Cheat bootstrapper is well under two seconds. Adoption dates a candidate
+    /// against the launcher's own start time rather than <see cref="EarlyExitThresholdSeconds"/>,
+    /// so this may be raised as far as a slow bootstrapper needs.
     /// </summary>
     public const int SpawnedChildDiscoveryTimeoutMs = 10_000;
 
--- a/GenHub/GenHub.Core/Helpers/GameProcessSelector.cs
+++ b/GenHub/GenHub.Core/Helpers/GameProcessSelector.cs
@@ -32,29 +32,85 @@ public static class GameProcessSelector
     }
 
     /// <summary>
-    /// Selects the process matching <paramref name="processName"/> that this launch spawned.
+    /// Selects the process matching <paramref name="processName"/> that this launch spawned, with
+    /// no launcher of ours to date the launch by — the storefront started the game itself. A
+    /// recency window is all that separates the new process from an instance of the same game that
+    /// was already running, so it is this path's only bound on age.
     /// </summary>
     /// <param name="candidates">The processes currently observed on the machine. Each candidate's <see cref="GameProcessCandidate.StartTime"/> must be a UTC <see cref="DateTime"/> with <see cref="DateTimeKind.Utc"/>.</param>
     /// <param name="processName">The expected process name, without extension.</param>
     /// <param name="workingDirectory">The directory the game must run from, or <see langword="null"/> to skip the check.</param>
     /// <param name="now">The current time, used to apply the recency window. Must be a UTC <see cref="DateTime"/> with <see cref="DateTimeKind.Utc"/>.</param>
-    /// <param name="launcherStartTime">The start time of the launcher process, if known. Must be a UTC <see cref="DateTime"/> with <see cref="DateTimeKind.Utc"/> when supplied.</param>
     /// <returns>The selected candidate, or <see langword="null"/> when none qualifies.</returns>
     public static GameProcessCandidate? SelectSpawnedGameProcess(
         IEnumerable<GameProcessCandidate> candidates,
         string processName,
         string? workingDirectory,
-        DateTime now,
-        DateTime? launcherStartTime = null)
+        DateTime now)
+    {
+        return Select(
+            candidates,
+            processName,
+            workingDirectory,
+            candidate => (now - candidate.StartTime).TotalSeconds < ProcessConstants.EarlyExitThresholdSeconds);
+    }
+
+    /// <summary>
+    /// Selects the process a launcher spawned, to be tracked and eventually terminated in the
+    /// launcher's place. Unlike <see cref="SelectSpawnedGameProcess"/> this refuses to answer at all
+    /// when the launcher's start time is unknown: without it, a process that started before this
+    /// launch and merely shares the name and the workspace cannot be told apart from the child, and
+    /// adopting it means killing somebody else's game when this launch is stopped.
+    /// <para>
+    /// That start time also replaces the recency window rather than joining it. It dates this
+    /// launch exactly, so anything at or after it started during the launch however long discovery
+    /// took, while a window measured against the clock expires a child that is genuinely ours the
+    /// moment the launcher is slow to produce it — and the discovery timeout the caller polls with
+    /// is configurable well past any fixed window. Keeping both would only turn a legitimate slow
+    /// adoption into an abandoned game that is still running.
+    /// </para>
+    /// </summary>
+    /// <param name="candidates">The processes currently observed on the machine. Each candidate's <see cref="GameProcessCandidate.StartTime"/> must be a UTC <see cref="DateTime"/> with <see cref="DateTimeKind.Utc"/>.</param>
+    /// <param name="processName">The expected process name, without extension.</param>
+    /// <param name="workingDirectory">The directory the game must run from, or <see langword="null"/> to skip the check.</param>
+    /// <param name="launcherStartTime">The start time of the launcher process. Must be a UTC <see cref="DateTime"/> with <see cref="DateTimeKind.Utc"/> when supplied.</param>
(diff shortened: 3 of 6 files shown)


bobtista:
All four findings were real and are fixed. One of them, followed literally, would have turned this PR into a no-op — details below, since that is the part worth reading.

### Truncated kernel process names

Confirmed on a purpose-built binary named `GeneralsOnlineZH_60`: `ProcessName == "GeneralsOnlineZ"` (exactly 15 chars), `GetProcessesByName` returns **0** for the full name and **1** for the 15-char prefix, while `MainModule.FileName` carries the full name. `GeneralsOnlineZH_60` is a real client name in this repo, so without this the PR did nothing for the clients it targets — and the existing harness child (`genhubchild`, 11 chars) could never have caught it.

Discovery now shortens the sought name to `ProcessConstants.UnixProcessNameMaxLength` on non-Windows only, and identity comes from the image path basename when readable (compared with and without extension, since Unix binaries carry no extension and may contain dots), falling back to `ProcessName` — exact or truncated — only when the path is unreadable. Windows is unaffected: there the basename and `ProcessName` always agree.

### Unknown launcher start time — and why the suggested fix would have broken this PR

The hazard is real, but it is **total on Unix, not occasional**, and I measured why: `Process.StartTime` throws `InvalidOperationException` for an exited process unless it was queried while the process was still alive, in which case .NET caches it. Verified directly — querying only after exit throws `Cannot process request because the process (…) has exited`, querying before exit and again after succeeds.

`HandleImmediateProcessExitAsync` runs *after* the launcher exits, so `launcherStartTime` was **always** null there. Applying the proposed change on its own — decline adoption when the start time is unknown — would therefore have disabled adoption on Unix entirely, silently reverting the behaviour this PR exists to add, while all tests still passed.

So the primary fix is to read the start time in `StartProcessAsync` immediately after `Process.Start`, while the launcher is alive (+0.1–1.7 ms, and in the same clock domain as candidate start times — verified a spawned grandchild compares correctly against it). The refusal guard is then a rare, safe backstop rather than the whole behaviour.

`AdoptExpectedChildProcessAsync` had the same exposure and is fixed the same way. It read the start time up front while the launcher was *usually* alive, so it was less often null — but "usually" is not a guarantee, and the consequence is identical: adopting and later killing a pre-existing instance of the same game from the same workspace. `FindSpawnedGameProcess` on the Steam discovery path keeps its unbounded policy, since it has no launcher to bound against.

### Residence: symlinks and case

Both fixed together, because fixing either alone is wrong. `ResidesIn` now compares with the shared `PathHelper.PathComparison`, and when that cheap compare fails it falls back to canonicalizing both paths component by component — resolving symlink targets (re-canonicalizing them, since a target may itself be reached through links, bounded by a new depth constant) and recovering each component's real on-disk spelling.

On the case question specifically: neither fixed rule is correct. `OrdinalIgnoreCase` everywhere can merge two genuinely distinct directories on Linux; `Ordinal` everywhere rejects a legitimate match on macOS, where a path differing only in case opens fine, launches fine, and then fails residence — reporting a running game as an immediate-exit failure. Component recovery uses .NET's `MatchCasing.PlatformDefault`, so both spellings canonicalize to the same real name on a case-insensitive volume and stay distinct on a case-sensitive one. Every step degrades to the caller's spelling on a missing or malformed path, so it cannot throw or abort the scan.

### Tests

Ten new pure `GameProcessSelector` unit tests, no new process spawning. Verified they bite: with image-path matching and the canonicalization fallback disabled, the truncated-name, symlink-residence and case-rule tests all fail. The two filesystem tests use real temp dirs with `finally` cleanup; the symlink one returns early where the platform refuses link creation, and the case test derives its expectation by probing the volume so it is correct on all three CI platforms.

### Noted, not changed

`SteamLauncher.StopRunningTargetProcessesAsync` calls `GetProcessesByName` with an untruncated name and has the same latent issue. Its failure mode is benign — it misses a kill rather than killing the wrong process — so I left it; it could adopt `GameProcessSelector.GetDiscoveryName` in a follow-up. `GenHub.ProxyLauncher` shares the pattern but is Windows-only, where the limit does not apply.

`LauncherHarness.Canonicalize` in the integration test is now redundant, since it existed to dodge the symlink issue. I left it deliberately: removing it would make a flaky real-process test the only guardian of the symlink fix.

1648 passing; the 3 failures are the environment-dependent real-engine tests, and I confirmed they produce byte-identical failures on the base commit with these changes reverted.

bobtista:
Two findings this round: one fixed, one refuted with a measurement.

### Fixed — the unknown-start-time case no longer polls to no purpose

`AdoptExpectedChildProcessAsync` now decides that case once, before the loop: a single error log, terminate the abandoned launcher, and return a failure. Terminating before building the message means the launcher has provably exited, so the full stderr is drained rather than a partial snapshot — the same ordering the timeout branch already uses. The `"Waiting up to {TimeoutMs}ms…"` info line moved below the guard so it is only emitted when a wait actually happens. Behaviour when the start time IS known is unchanged: the loop, deadline, grace period and all four exit branches are untouched.

On the same comment's secondary claim, that the timeout branch "kills a launcher that may really have spawned the game" — that branch is unchanged and the new fail-fast path also terminates the launcher, deliberately. Returning a failure while leaving an untracked bootstrapper, and possibly a running game, behind is worse than the kill.

**Audit of every other per-iteration log in that loop**, since the finding was really about a shape rather than one line:
- `FindGameProcess`'s catch blocks log a warning and return null, which does repeat every 100 ms if `Process.GetProcessesByName` fails persistently. **Deliberately left polling**: unlike a missing start time, an enumeration failure is not provably permanent — a transient permission or handle failure can recover — so giving up would be wrong. It is the one remaining candidate if the team wants to revisit it.
- `ReadLauncherExit` and `GetProcessExecutablePath` log at Debug by documented policy; the `EnableRaisingEvents` warning and all failure branches fire at most once.

### Refuted — `OnDiskName` case folding

The CRITICAL states that .NET `searchPattern` matching is case-sensitive on every Unix platform, so `OnDiskName` cannot fold case and the residence check will reject a legitimate match on macOS.

I measured it with the exact call the code makes, on macOS, in a directory containing `MyWorkspace`:

```
Directory.GetFileSystemEntries(parent, "myworkspace") -> 1 hit(s)
   MyWorkspace
```

So case folding does work through that overload on macOS. On Linux the match is case-sensitive, which is the correct behaviour for a case-sensitive volume — that asymmetry is the intent, not a defect, and it is why the code relies on the platform's own rules instead of a fixed `StringComparison`. No change made.

### On the missing test

I deliberately did not add one. Reaching the new branch requires `Process.StartTime` to throw on a live process; `ReadStartTime` is private, the class exposes no virtual or protected members, and `StartProcessAsync` only accepts a real executable path. Covering it honestly would mean adding a `protected virtual` seam to production code purely for the test plus a spawn-based timing test. The cheap alternative — a helper asserting `HasValue` — would only restate the guard. Happy to add the seam and the spawn-based test if the team would rather have it.

1648 passing; the 3 failures are the environment-dependent real-engine tests.

undead2146 (dismissed):
LGTM

bobtista:
Fixed by removing the bound rather than adjusting it.

### Why the window went away on the adoption path

`SelectAdoptableGameProcess` applied **both** bounds: the 10-second recency window and `StartTime >= launcherStartTime`. The polling loop runs for a configurable `ExpectedChildDiscoveryTimeout` that can exceed 10 s, so a child that only became enumerable after that point was rejected on every poll until the loop reported a discovery timeout — killing the launcher and failing a launch whose game was actually running.

The recency window is a **proxy** for "this process belongs to this launch", used where nothing better exists. The launcher start time is a direct measurement of the same thing: we spawned that launcher at the start of this launch, so any candidate at or after it started during this launch, however long discovery took. On the adoption path the window is therefore not merely redundant but strictly harmful — it can only reject a candidate that is genuinely ours, never accept one that is not. So it is dropped there rather than widened, since widening only moves the cliff and the timeout is caller-configurable with no upper bound worth encoding. The reviewer was right that clamping would only shorten the failure; with the window gone there is nothing left to clamp against.

### Made the two policies impossible to confuse

The shared name/residence/newest-first logic moved into a private helper, and each public entry point now supplies its own "belongs to this launch" predicate:

- `SelectSpawnedGameProcess` (Steam discovery, no launcher to bound against) keeps the recency window as its only bound and **lost** its unused optional `launcherStartTime` parameter — production never passed one. A dead optional parameter on that path is exactly the shape that let the two policies get conflated, so removing it makes weakening that path structurally impossible rather than merely discouraged.
- `SelectAdoptableGameProcess` keeps the outright refusal when the launcher start time is unknown and keeps `StartTime >= launcherStartTime`, so a candidate predating the launcher is still rejected. It **lost** its `now` parameter and no longer consults a clock at all, which is the property the fix rests on.

`ProcessConstants.SpawnedChildDiscoveryTimeoutMs` no longer documents a "must not exceed `EarlyExitThresholdSeconds`" coupling — that coupling was the finding, and it no longer holds.

### Tests

`SelectAdoptableGameProcess_AdoptsAChildOlderThanTheRecencyWindow` — launcher started well beyond the window, child 1 s after it, must be adopted. Verified failing against the pre-fix code. The pre-existing rejection tests for a candidate predating the launcher, an unknown start time, and the Steam path expiring stale candidates all still pass.

That new test is anchored to `DateTime.UtcNow` rather than the class's frozen clock, deliberately: since the adoption path no longer takes a time parameter, a reintroduced window would have to read the ambient clock, and this anchoring catches that. Confirmed by temporarily re-adding the window — all three adoption tests failed, then passed on revert.

1651 passing.

### One judgement call

Removing parameters from two public helpers in `GenHub.Core` is source-breaking. `GameProcessSelector` is internal-use-only in this repo and every call site is updated. The alternative was leaving a dead optional parameter on the Steam path, which is the thing that caused this.

undead2146 (approved):
LGTM

PR #467, opened 2026-09-12
fix(ui): keep dialog markdown readable in light system mode

### LINKED ISSUE #466: Getting Started text is unreadable with Windows light app mode
The Getting Started dialog renders black body text on a dark background when Windows uses light app mode, making the introduction and bullets difficult to read.

**Repro:** Use Windows light app mode and launch GenHub with the first-run dialog enabled. Expected: readable light text on the dialog's dark background.

**Cause:** The dialog styles `TextBlock`, but Markdown.Avalonia renders `CTextBlock`. Its foreground follows the system theme while the background stays dark. Reproduced under Avalonia Light and Dark themes.

**Fix:** Set the markdown viewer's inherited `TextElement.Foreground` to `{DynamicResource TextPrimaryBrush}` and cover both themes with a regression test.

Found while validating #378.

bobtista:
Fixes #466.

Getting Started used unreadable markdown colors when Windows selected Light mode. Use `TextPrimaryBrush` so dialog markdown remains readable in both themes.

Validation: 6 targeted Light/Dark tests pass.


undead2146 (dismissed):
Dont think we need the docs/reviews/dialog-markdown-contrast/after.png  and docs/reviews/dialog-markdown-contrast/before.png images. 
besides that LGTM

PR #469, opened 2026-09-12
fix(catalog): exclude Launcher.txt from Generals registry

bobtista:
Remove the optional `Launcher.txt` documentation file from the Generals catalog and refresh the catalog metadata and pinned checksum.

**Testing:** Catalog and CSV resolver tests pass.


undead2146 (approved):
LGTM

PR #475, opened 2026-09-12
fix(gameprofiles): persist the active workspace id prepared by a profile's own launch

### LINKED ISSUE #460: ActiveWorkspaceId is not persisted after launching a profile
## Problem

`GameLauncher` registers the launch before `ProfileLauncherFacade` attempts to persist the prepared workspace ID.

`GameProfileManager.UpdateProfileAsync` then considers the profile running and rejects the update with:

> Cannot change active workspace while profile is running.

The launch succeeds, but `ActiveWorkspaceId` remains empty. Content reconciliation can therefore fail to locate and invalidate the active workspace later.

## Acceptance criteria

- Persist the prepared workspace ID during launch before the running-profile guard can reject it.
- Do not weaken the guard against user-initiated workspace changes while a game is running.
- Do not report launch success if required workspace state cannot be persisted safely.
- Add a regression test covering a successful launch from an empty `ActiveWorkspaceId`.

bobtista:
Closes #460.

## Problem

`GameLauncher` registers a placeholder launch before it prepares the workspace, so the profile already counts as running by the time `ProfileLauncherFacade` persists the prepared workspace id. `ValidateRunningProfileImmutableSettings` then rejects the update with "Cannot change active workspace while profile is running."

The launch succeeds but `ActiveWorkspaceId` stays empty, so `ContentReconciliationService` cannot find and invalidate the workspace later.

Observed on a real launch:

```
[GameLauncher] Process started successfully - PID: 22508
Registered launch 825eafc1-... for profile c1afa642...
=== LAUNCH SUCCESS: Profile c1afa642..., ProcessId 22508 ===
[WRN] Failed to persist active workspace ID to profile 'c1afa642...':
      Cannot change active workspace while profile is running.
```

## Change

The guard now distinguishes a profile recording the workspace its **own active launch** prepared from a user pointing the profile at a **different** workspace. The first is reconciliation and is allowed. The second is still rejected.

`GameProfileManager` reads the active launch's `WorkspaceId` from `ILaunchRegistry` and passes it into the immutable-settings check. By the time the facade persists, the registry already holds the real workspace id, so the match succeeds.

The guard is not weakened: every other immutable setting is untouched, and an update naming a workspace other than the running one still fails.

## Tests

Two regression tests in `GameProfileManagerHotswapTests`:

- `UpdateProfileAsync_WhenRunning_PersistsActiveWorkspaceIdFromOwnLaunchAsync` covers launch from an empty `ActiveWorkspaceId`.
- `UpdateProfileAsync_WhenRunning_RejectsDifferentActiveWorkspaceIdAsync` confirms the guard still blocks user-initiated changes and saves nothing.

Verified the first test genuinely catches the defect. With the production change reverted:

```
UpdateProfileAsync_WhenRunning_PersistsActiveWorkspaceIdFromOwnLaunchAsync [FAIL]
Failed! - Failed: 1, Passed: 11, Total: 12
```

With the change applied:

```
Passed! - Failed: 0, Passed: 12, Total: 12
```

Release build of `GenHub.csproj` reports no warnings.

## Note on the third acceptance criterion

"Do not report launch success if required workspace state cannot be persisted safely" is not addressed here. After this change the persist succeeds, so the case no longer arises on the normal path. Turning a late persistence failure into a launch failure while the game is already running looked like a worse outcome than surfacing it, so I left that decision open rather than guess.

### DIFF
--- a/GenHub/GenHub.Tests/GenHub.Tests.Core/Features/GameProfiles/Services/GameProfileManagerHotswapTests.cs
+++ b/GenHub/GenHub.Tests/GenHub.Tests.Core/Features/GameProfiles/Services/GameProfileManagerHotswapTests.cs
@@ -529,6 +529,75 @@ public class GameProfileManagerHotswapTests
         Assert.Equal("-restoredargs", rollbackResult.Data?.CommandLineArguments);
     }
 
+    /// <summary>
+    /// Verifies that a running profile accepts the workspace id its own active launch prepared.
+    /// The launch path persists this after the placeholder launch has already been registered,
+    /// so the running-profile guard must not reject it.
+    /// </summary>
+    /// <returns>A task representing the test operation.</returns>
+    [Fact]
+    public async Task UpdateProfileAsync_WhenRunning_PersistsActiveWorkspaceIdFromOwnLaunchAsync()
+    {
+        // Arrange
+        const string profileId = "profile-launch";
+        const string workspaceId = "ws-launched";
+        var existingProfile = new GameProfile
+        {
+            Id = profileId,
+            Name = "Launching Profile",
+            ActiveWorkspaceId = string.Empty,
+        };
+
+        _profileRepositoryMock.Setup(r => r.LoadProfileAsync(profileId, It.IsAny<CancellationToken>()))
+            .ReturnsAsync(ProfileOperationResult<GameProfile>.CreateSuccess(existingProfile));
+        _profileRepositoryMock.Setup(r => r.SaveProfileAsync(It.IsAny<GameProfile>(), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(ProfileOperationResult<GameProfile>.CreateSuccess(existingProfile));
+        _launchRegistryMock.Setup(l => l.GetAllActiveLaunchesAsync())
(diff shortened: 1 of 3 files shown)


bobtista:
Addressing the reviewer note on the third acceptance criterion, "do not report launch success if required workspace state cannot be persisted safely".

I looked at making the facade return a failure when `UpdateProfileAsync` fails, and it is the wrong move here. The persist happens after the process has started, so a failure result would flow to `GameProfileLauncherViewModel.LaunchProfileAsync`, which only sets `IsProcessRunning = true` on the success branch. The game would be running while GenHub believed it was not, which means no Stop button and no exit tracking for a live process. That trades a logged inconsistency for an untracked one.

What this PR does instead:

- Removes the cause. The guard was rejecting the profile's own launch, so the normal path now persists successfully. That is the actual bug in #460.
- Makes the residual failure loud. The `updateResult.Success == false` branch was `LogWarning`; it is now `LogError` and names the consequence, that content reconciliation cannot invalidate the workspace later.

Failing the launch after a successful spawn needs a rollback story, terminate the process and clean the workspace, which is a larger behavioural change than this fix and should be its own issue if wanted. Happy to open one.

bobtista:
@coderabbitai review

undead2146 (approved):
LGTM

PR #546, opened 2026-09-20
fix(downloads): correct release state matching, variant grouping, and progress fan-out

undead2146:
## Summary
Fixes downloads/library state correctness and UI responsiveness: ModDB release rows now link to their own manifests across restarts, GeneralsOnline release manifests collapse into single library cards, release-row manifest binding requires exact identity, and download progress fan-out is throttled so large archives cannot freeze the UI.

### Root Cause
Several independent defects in the downloads pipeline: file-row lookup never consulted the per-file detail-URL provenance both sides share, so ModDB rows lost their Downloaded state after restart; GeneralsOnline's client, game-data patch, and mappack listed as three separate cards; loose name/version matching bound "Add to Profile" to a sibling release's manifest; per-file extraction progress flooded the UI thread; the detail view shared one mutable search result with the card; and ModDB downloads stacked 3-4 toasts.

### Changes
- **ContentStateService**: file-row lookup links rows to manifests via shared detail URL (gated on name/version agreement); version heuristics no longer veto exact provenance linkage for file rows.
- **GeneralsOnline**: new `GeneralsOnlineVariantGrouping` groups one release's manifests by full version (QFE included); factory stamps group id/family name; discoverer derives grouping for legacy pool entries and maps library pages defensively.
- **ContentDetailViewModel**: release rows flip state and bind manifests only on exact manifest/name identity (plus exact version match when fully downloaded with no update pending); removed the stacking pre-download ModDB toast.
- **DownloadsBrowserViewModel / VariantSwap**: detail view receives a snapshot of the card's search result; variant swaps preserve aggregated page data/data payloads when the variant carries none.
- **ContentDownloadCoordinator / Core**: new `DownloadProgressThrottle` (150 ms, phase changes and completion bypass) for progress fan-out; staging cleanup moved off the calling thread; per-file hashing log demoted to Debug; Playwright skips the duplicate per-download toast when it already announced the browser launch.
- **Tests**: ModDB row/manifest linkage tests, GO variant grouping tests, throttle unit tests, variant-swap preservation tests, library grouping/defensive-mapping tests; CloseDetail mock matches by catalog id instead of reference identity.

### Verification
- [x] Targeted unit/integration tests executed and passing (5714 passed; 2 remaining failures verified pre-existing on the clean tree: a locale-dependent quota-formatting test and a CsvResolver cancellation test)
- [x] Solution/project builds cleanly without new warnings or lint errors
- [x] Verified cross-platform compatibility where applicable (no OS-specific hooks touched)

---
*Created with Muse Spark via Muse Code*


undead2146:
@coderabbitai review

undead2146:
@coderabbitai review

